PR #298, opened 2018-04-23
feat(judge): implement NaN replacement strategy

skandragon:
This change makes it possible to specify a NaN replacement strategy, in addition to the (default) removal policy we now have.  It will replace with 0, which is currently the only sane value to replace it with.

### DIFF
--- a/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
+++ b/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
@@ -26,7 +26,7 @@ case object Nodata extends MetricClassificationLabel
 case object Error extends MetricClassificationLabel
 
 sealed trait MetricDirection
-object MetricDirection{
+object MetricDirection {
   case object Increase extends MetricDirection
   case object Decrease extends MetricDirection
   case object Either extends MetricDirection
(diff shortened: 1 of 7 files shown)


kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
```diff
@@ -41,6 +41,20 @@ object MetricDirection{
   }
 }
 
+sealed trait NanStrategy
+object NanStrategy {
```
csanden: The classifiers do not use the NanStrategy directly. Towards this end, should this object be defined in the BaseMeticClassifier or should it be defined elsewhere?

kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
```diff
@@ -41,6 +41,20 @@ object MetricDirection{
   }
 }
 
+sealed trait NanStrategy
```
csanden: For consistency with other projects (Apache Math) this should be defined as `NaNStrategy`

kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
skandragon: I think the whole thing needs to be refactored as proper objects; when we do that, I think we'll end up with the configuration part moving outside both, and called from either as needed.  For now, let's leave it where it is, since it sits nicely next to the other config options.

kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/BaseMetricClassifier.scala
skandragon: Will fix.  I also added a test (gasp!)

PR #300, opened 2018-04-24
Adding docker-compose support, run instructions in README and minor fixes to broken external dependencies

flaviostutz:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8
+FROM openjdk:8-jdk
```
lwander: Why is the jdk needed?

Dockerfile
```diff
@@ -11,4 +11,8 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
   cd .. && \
   rm -rf workdir
 
+EXPOSE 8090
+ENV services.redis.baseUrl=redis://localhost:6379
+ENV redis.connection=redis://localhost:6379
```
lwander: These env & port settings should either be defaults in the mounted configuration, or overridden in a kubernetes or docker compose config.

gradle.properties
```diff
@@ -1,5 +1,5 @@
 springVersion=4.3.10.RELEASE
-springBootVersion=1.5.10.RELEASE
+springBootVersion=1.5.12.RELEASE
```
lwander: Is this upgrade needed for docker compose to work?

Dockerfile
flaviostutz: The non-jdk version was not working (broken pipe when accessing Docker Hub). The JDK version builds successfully. Maybe it was temporary, but until today it doesn't work. Could you test in your machine?

Dockerfile
flaviostutz: It is a good practice to declare in Dockerfile all exposed PORTs and ENVs. To do that you have to define a default value, even one that will be overwritten by docker-compose.

gradle.properties
flaviostutz: It is needed because sprintboot 1.5.10 doesn't exist in MVN repo anymore.

Dockerfile
lwander: What about `-jre`?

flaviostutz: According to https://hub.docker.com/r/library/openjdk/tags/, this tag doesn't exist

Dockerfile
flaviostutz: If those configuration settings are meant to be configured through Docker ENV parameters, you should list them in Dockerfile as a way for developers to know they exist!

Dockerfile
duftler: I just provisioned a new vm, installed docker, and it built just fine.
The top one is the new kayenta image and the bottom one is the `openjdk:8` image.

```
duftler$ sudo docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
<none>              <none>              e9eece7d9b97        8 seconds ago       900.3 MB
openjdk             8                   c3e386dcd062        4 days ago          726.1 MB
```

gradle.properties
duftler: It's possible I'm looking in the wrong place, but I see these:
https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-gradle-plugin/1.5.10.RELEASE
https://mvnrepository.com/artifact/org.springframework.boot/spring-boot/1.5.10.RELEASE

Also, the other Spinnaker services all seem to still be using `1.5.10` (or earlier) and I'm uncomfortable moving just this one to a later version with a compelling reason (I checked a handful of the others).

Lastly, it builds just fine for me, for a bunch of others, and in our nightly builds, so it seems like there isn't an issue resolving the currently-referenced version.

skandragon:
This PR has been open for a while now.  Can we make a decision to update it (as it does not merge and changes were requested) or reject it?

duftler:
Going to close this one, since it's stale. Please reopen if still relevant.

PR #326, opened 2018-06-19
fix(Judge): Handle Nodata groups correctly

skandragon:
New nodata rules:

* If 50% or more metrics in the entire canary report Nodata, the final score will be forced to 0.0.
* If all metrics inside a group report Nodata, the group is marked with a 0.0 score.
* Compute the summary score as before, except that entirely Nodata groups are considered a 100% Pass, which in effect makes its weight ignored.  That is, if we have two groups, one with just Nodata, one with a score of 100.0, the final result will be 100.0.


skandragon:
Self-review:  needs work.  :)

Issue #333, opened 2018-06-26
Atlas NaN replacement does not handle completely empty responses from atlas


skandragon:
If a query results in no data at all, we do not properly pad the query and replace NaNs.

I feel we should always pad the query...

skandragon:
Closed by #335

PR #335, opened 2018-06-26
fix(atlas): create an empty array for no results

skandragon:
When Atlas returns no data, we previously returned an empty list.  Now, we create a list of `NaN` values exactly sized for the number of data points we expected to receive.


skandragon:
Also, fixes #333

skandragon:
This is more complicated than I expected.  Working on it still.

Issue #332, opened 2018-06-26
Support for stackdriver resource type: gke_container


ariefrahmansyah:
Hi, will kayenta add support for Stackdriver's `gke_container` resource type? https://cloud.google.com/monitoring/api/resources#tag_gke_container

We still use old version of Stackdriver Kubernetes monitoring and still don't have plan to migrating to new version of Kubernetes monitoring (https://cloud.google.com/monitoring/kubernetes-engine/migration). It will help us a lot if we can support `gke_container`.

Thanks.

duftler:
UI support was added here: https://github.com/spinnaker/deck-kayenta/pull/315

PR #337, opened 2018-06-27
feat(stackdriver): Add support for legacy gke_container resource type.

duftler:


duftler:
Closes https://github.com/spinnaker/kayenta/issues/332.

Issue #301, opened 2018-05-03
Error Handling - Tied values fail canary analysis


e96wic:
I created an example metric that measures my container memory limit. If this does not change in one interval (which the limit usually does not), I get an error which is caused by [this](https://github.com/spinnaker/kayenta/blob/f704192fa8fb7beaf75d40c74ab85c680a404d33/kayenta-mannwhitney/src/main/scala/com/netflix/kayenta/mannwhitney/MannWhitney.scala#L93).

This is probably some kind of error handling problem - I would expect a proper result of a canary analysis using constant values.

skandragon:
We actually discovered this recently as well, and have a fix planned.

csanden:
This issue was fixed in #343.

PR #343, opened 2018-07-04
fix(judge): Handle tied ranks by adding Gaussian white noise

csanden:
Handle the scenario (degenerate distribution) where the ranks, as computed by Mann-Whitney, are tied resulting in an `Error` classification as discussed in #301 . 

To address this, we add a small amount of [Gaussian white noise](https://en.wikipedia.org/wiki/White_noise) to the input signal when both the experiment and control signals are constant.

### DIFF
--- a/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/MannWhitneyClassifier.scala
+++ b/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/MannWhitneyClassifier.scala
@@ -17,6 +17,7 @@
 package com.netflix.kayenta.judge.classifiers.metric
 
 import com.netflix.kayenta.judge.Metric
+import com.netflix.kayenta.judge.preprocessing.Transforms
 import com.netflix.kayenta.mannwhitney.{MannWhitney, MannWhitneyParams}
 import org.apache.commons.math3.stat.StatUtils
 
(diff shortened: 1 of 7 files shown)


kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/MannWhitneyClassifier.scala
```diff
@@ -55,6 +57,15 @@ class MannWhitneyClassifier(tolerance: Double=0.25, confLevel: Double=0.95) exte
     (lowerBound, upperBound)
   }
 
+  /**
+    * Add Gaussian noise to the input array
+    */
+  private def addGaussianNoise(values: Array[Double]): Array[Double] = {
+    val scalingFactor = 1e-5
+    val metricScale = values.distinct.head * scalingFactor
```
lwander: what happens if the first value is 0 or NaN?

csanden: If this happens then the `metricScale` would be 0 and NaN respectively. Note that, NaNs are transformed, either removed or replaced, before the classifier is called. In spite of this, I will add some guards to check for the zero or NaN case.

skandragon:
Fixes #301

skandragon:
Chris and I discussed, and he's going to see if we can fix the 0-case by just using a constant SD for the noise that is always very small.

I recommended addressing the NaN case by documentation:  "this method requires there be no NaNs in the input array".

csanden:
After further discussion we have decided to scale the amplitude of the Gaussian noise based on the input values. Note that, NaN values should be removed or replaced prior to classification. Furthermore, zero value inputs are handled correctly as both inputs (experiment and control) are checked prior to classification.

PR #351, opened 2018-07-14
fix(judge): Round the Summary Score to 2 decimal places

### LINKED ISSUE #271: Handling of floating point scores is sometimes surprising
I ran a canary where the weighted sum resulted in a 5 (1 out of 3 passing metrics and the group has 15% as it's weight).

But the ui shows a 4 as the score, so I looked at the result returned from kayenta and it has this:
```
"score": {
  "classification": "Fail",
  "classificationReason": "",
  "score": 4.999999999999999
}
```

I understand we are probably truncating/flooring to avoid overplaying the confidence, but in this case, 1/3 of 15 is 5 and I think a user would be surprised to see anything but 5. Seems like rounding (maybe within some threshold) would give a result closer to what a user would expect.

That is, maybe it wouldn't make sense to round 4.6 up to 5, but 4.999999999999999 should be rounded to 5 (esp. since it's actually supposed to be 5).

csanden:
Round the final Summary Score to 2 decimals places (Fixes #271).

### DIFF
--- a/kayenta-judge/src/test/scala/com/netflix/kayenta/judge/ScorerSuite.scala
+++ b/kayenta-judge/src/test/scala/com/netflix/kayenta/judge/ScorerSuite.scala
@@ -17,10 +17,9 @@
 package com.netflix.kayenta.judge
 
 import com.netflix.kayenta.canary.results.CanaryAnalysisResult
+import com.netflix.kayenta.judge.classifiers.metric.{High, Low, Nodata, Pass}
 import com.netflix.kayenta.judge.scorers.WeightedSumScorer
-import com.netflix.kayenta.judge.classifiers.metric.{Pass, High, Low, Nodata}
 import org.scalatest.FunSuite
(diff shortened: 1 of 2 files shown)


kayenta-judge/src/test/scala/com/netflix/kayenta/judge/ScorerSuite.scala
```diff
@@ -98,7 +97,7 @@ class ScorerSuite extends FunSuite {
 
     assert(scores.groupScores.isDefined)
     val groupScores = scores.groupScores.get
-    assert(groupScores.size == 2)
+    assert(groupScores.lengthCompare(2) == 0)
```
skandragon: This change seems less readable.  Was it made for a specific problem or was this an IDE suggestion?

csanden: This was an IDE suggestion.

Issue #360, opened 2018-08-06
Add MetricsInterceptors


duftler:
The other services all have this:
Rosco -> https://github.com/spinnaker/rosco/blob/master/rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/WebConfig.groovy#L33
Clouddriver -> https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/WebConfig.groovy#L58

PR #367, opened 2018-08-22
chore(monitoring): Record controller invocations.

duftler:


kayenta-web/src/main/java/com/netflix/kayenta/interceptors/MetricsInterceptor.java
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright 2015 Netflix, Inc.
```
duftler: Left this as I simply copied [this](https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/interceptors/MetricsInterceptor.java) and replaced the path variable lookup with a query param lookup.

duftler:
Closes #360.

kayenta-web/src/main/java/com/netflix/kayenta/interceptors/MetricsInterceptor.java
ewiseblatt: mention this is copied from the kato url

duftler: Done.

kayenta-web/src/main/java/com/netflix/kayenta/interceptors/MetricsInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      Id id = registry.createId(metricName)
+        .withTag("controller", controller)
+        .withTag("method", handlerMethod.getMethod().getName())
+        .withTag("status", status.toString().charAt(0) + "xx")
+        .withTag("statusCode", status.toString());
+
+      List<String> paramNames = Collections.list(request.getParameterNames());
+
```
ewiseblatt: Can you mark this as the place that differs, and perhaps todo generalizing kork in such a way as to support this use case so you dont need this variant?

Leave for now but note it.

duftler: Done.

PR #374, opened 2018-09-12
feat(cors-headers): add if we have an Origin header on request

skandragon:
This will echo back the contents of the `Origin` header, if present, and add additional CORS headers.  No checking is done on the origin; this could be added later if needed.

### DIFF
--- a/kayenta-web/src/main/java/com/netflix/kayenta/filters/KayentaCorsFilter.java
+++ b/kayenta-web/src/main/java/com/netflix/kayenta/filters/KayentaCorsFilter.java
@@ -1,0 +1,31 @@
+package com.netflix.kayenta.filters;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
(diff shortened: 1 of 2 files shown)


kayenta-web/src/main/java/com/netflix/kayenta/filters/KayentaCorsFilter.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public void init(FilterConfig filterConfig) throws ServletException {}
+
+  @Override
+  public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+    HttpServletResponse response = (HttpServletResponse) res;
+    HttpServletRequest request = (HttpServletRequest) req;
+    String origin = request.getHeader("Origin");
+    if (origin != null) {
```
duftler: Could `origin` be an empty string?

skandragon: Hmm, I will add a check for that as well, and push it.  I'm not sure what would happen there, but 
I suspect it would "just work" as well as anything else.

PR #375, opened 2018-09-12
feat(locations): provide location hints for atlas

skandragon:
This is a partial fix for some Atlas complexity reduction.

This adds an ability to provide UI hints about locations.  For Atlas, Netflix has many deployment, environment, and datacenter combinations to choose from.  We have started changing to a model where these magic location strings will be sufficient to provide all these details, as they are unique across all our atlas deployments.

With this change, most of the special Atlas scope parameters will be replaced by the single location, which will greatly simplify the code.

Optionally, a recommended list may be provided in the kayenta config YAML file. If provided, it will be inserted into the `recommendedLocations` list; otherwise, the list should be empty.

I recommend the UI to present a text input box if `locations` is empty, otherwise present a pull-down with the `recommendedLocations` first, with options to switch to `locations` for the full list.  For Atlas, it makes little sense to allow arbitrary text input here, but it may also be useful for other metric sources to have a text field option as well even with locations provided if they cannot all be discovered.

### DIFF
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
@@ -41,7 +41,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import javax.validation.constraints.NotNull;
-import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
@@ -81,11 +80,17 @@ public class AtlasMetricsService implements MetricsService {
     return accountNames.contains(accountName);
   }
 
+  private AtlasNamedAccountCredentials getCredentials(String accountName) {
+    return (AtlasNamedAccountCredentials)accountCredentialsRepository
+      .getOne(accountName)
+      .orElseThrow(() -> new IllegalArgumentException("Unable to resolve account " + accountName + "."));
+  }
+
   @Override
   public List<MetricSet> queryMetrics(String accountName,
                                       CanaryConfig canaryConfig,
                                       CanaryMetricConfig canaryMetricConfig,
-                                      CanaryScope canaryScope) throws IOException {
+                                      CanaryScope canaryScope) {
 
     OkHttpClient okHttpClient = new OkHttpClient();
     // TODO: (mgraff, duftler) -- we should find out the defaults, and if too small, make this reasonable / configurable
@@ -99,9 +104,7 @@ public class AtlasMetricsService implements MetricsService {
     }
 
     AtlasCanaryScope atlasCanaryScope = (AtlasCanaryScope)canaryScope;
-    AtlasNamedAccountCredentials credentials = (AtlasNamedAccountCredentials)accountCredentialsRepository
-      .getOne(accountName)
-      .orElseThrow(() -> new IllegalArgumentException("Unable to resolve account " + accountName + "."));
+    AtlasNamedAccountCredentials credentials = getCredentials(accountName);
     Optional<Backend> backend = credentials.getBackendUpdater().getBackendDatabase().getOne(atlasCanaryScope.getDeployment(),
                                                                                             atlasCanaryScope.getDataset(),
                                                                                             atlasCanaryScope.getLocation(),
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
@@ -34,7 +34,9 @@ package com.netflix.kayenta.atlas.model;
 import lombok.*;
 
 import javax.validation.constraints.NotNull;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 @Builder
 @ToString
@@ -46,6 +48,10 @@ public class Backend {
   @NotNull
   @Getter
   private String cname;
+
+  @NotNull
+  @Getter
+  private String target;
 
   @NotNull
   @Getter
@@ -72,12 +78,41 @@ public class Backend {
   @Getter
   private String description;
 
-  public String getUri(String method, String deployment, String dataset, String region, String environment) {
-    String ret = cname
+  public List<String> getTargets() {
+    ArrayList<String> ret = new ArrayList<>();
+
+    ret.add(target.replace("$(deployment)", deployment).replace("$(dataset)", dataset));
+
+    if (regions != null) {
+      ArrayList<String> newList = new ArrayList<>();
+      for (String region : regions) {
+        for (String item : ret) {
+          newList.add(item.replace("$(region)", region));
+        }
+      }
+      ret = newList;
+    }
+
+    if (environments != null) {
+      ArrayList<String> newList = new ArrayList<>();
+      for (String environment : environments) {
+        for (String item : ret) {
+          newList.add(item.replace("$(env)", environment));
+        }
+      }
+      ret = newList;
+    }
+
+    return ret.stream().distinct().collect(Collectors.toList());
+  }
+
+  public String getUri(String scheme, String deployment, String dataset, String region, String environment) {
+    String relativeReference = cname
       .replace("$(deployment)", deployment)
       .replace("$(dataset)", dataset)
       .replace("$(region)", region)
       .replace("$(env)", environment);
-    return method + "://" + ret;
+
+    return scheme + "://" + relativeReference;
   }
 }
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/security/AtlasNamedAccountCredentials.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/security/AtlasNamedAccountCredentials.java
@@ -24,6 +24,7 @@ import lombok.Data;
 import lombok.Singular;
 
 import javax.validation.constraints.NotNull;
+import java.util.Collections;
 import java.util.List;
 
 @Builder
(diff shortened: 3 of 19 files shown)


kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
```diff
@@ -81,11 +81,34 @@ public boolean servicesAccount(String accountName) {
...
+    return credentials.getBackendUpdater().getBackendDatabase().getLocations();
+  }
+
+  @Override
+  public List<String> getRecommendedLocations(String accountName) {
+    AtlasNamedAccountCredentials credentials = (AtlasNamedAccountCredentials)accountCredentialsRepository
+      .getOne(accountName)
+      .orElseThrow(() -> new IllegalArgumentException("Unable to resolve account " + accountName + "."));
+
+    if (credentials.getRecommendedLocations() == null) {
```
duftler: The if/else here are in the opposite order of the way the logic was explained in the commit comment. I think it reads more naturally as described in the comment (if it's present, use it, otherwise fall back to using `locations` instead).

Also, this will not fall back to `locations` if `recommendedLocations` is provided but is an empty list. I think this is what you intended, but wanted to check just in case.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
```diff
@@ -72,12 +78,41 @@
   @Getter
   private String description;
 
+  public List<String> getTargets() {
```
duftler: Might be worth a sentence or two for future readers to know what is going on in this method.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
```diff
@@ -72,12 +78,41 @@
...
   public String getUri(String method, String deployment, String dataset, String region, String environment) {
-    String ret = cname
+    String uri = cname
       .replace("$(deployment)", deployment)
       .replace("$(dataset)", dataset)
       .replace("$(region)", region)
       .replace("$(env)", environment);
-    return method + "://" + ret;
+
+    return method + "://" + uri;
```
duftler: Is this `method` and not `scheme`?

I don't feel strongly about it, but I expected `uri` to contain everything, including the scheme (that is, I don't think it's a uri until the scheme is appended to the front; but I could be wrong).

Maybe s/uri/relativeReference and `return scheme + "://" + relativeReference`.

Or just leave it as is...

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/security/AtlasNamedAccountCredentials.java
```diff
@@ -37,16 +37,19 @@
...
 
-  @Override
-  public String getType() {
-    return "atlas";
-  }
+  private List<String> recommendedLocations;
 
   @JsonIgnore
   private BackendUpdater backendUpdater;
+
```
duftler: Is this extra blank line intentional? Have seen this with some of the `.scala` files, so maybe it's just how you auto-format.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourcesController.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2017 Google, Inc.
```
duftler: s/2017 Google/2018 Netflix

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourceDetail.java
```diff
@@ -0,0 +1,29 @@
+package com.netflix.kayenta.controllers;
```
duftler: Missing license header.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsService.java
```diff
@@ -38,4 +38,9 @@
   default List<Map> getMetadata(String metricsAccountName, String filter) throws IOException {
     return Collections.emptyList();
   }
+
+  List<String> getLocations(String metricsAccountName);
```
duftler: Could use default methods for both of these so you won't need to touch any of the implementing classes other than `AtlasMetricsService`.

duftler: Like `getMetadata()` just above.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsService.java
```diff
@@ -38,4 +38,9 @@
...
+
+  List<String> getRecommendedLocations(String metricsAccountName);
```
duftler: Comments on both of these methods would also be helpful since they are only used by Atlas at the moment (esp. for implementors of new metrics service integrations).

kayenta-core/src/main/java/com/netflix/kayenta/security/MetricAccountCredentials.java
```diff
@@ -0,0 +1,9 @@
+package com.netflix.kayenta.security;
+
+import java.util.List;
+
+public interface MetricAccountCredentials<T> extends AccountCredentials<T> {
```
duftler: I don't see this interface used anywhere (nor do I see where it would be, so I think it extraneous).

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourcesController.java
```diff
@@ -0,0 +1,72 @@
...
+        MetricsService metricsService =
+          metricsServiceRepository
+            .getOne(account.getName())
+            .orElseThrow(() -> new IllegalArgumentException("No metrics service was configured; unable to read from metrics store."));
+
+        return MetricSourceDetail
+          .builder()
+          .name(account.getName())
+          .type(metricsService.getType())
+          .locations(metricsService.getLocations(account.getName()))
```
duftler: Since the logic in `metricsService.getLocations` and `metricsService. getRecommendedLocations` operate only on the credentials objects, I can't see why this needs to go through the metrics service at all? Would all be much simpler if the `locations` and `recommendedLocations` properties were only exported from the credentials class, and not from the metrics service class.

duftler: Naming aside, the `CredentialsController` already returns all configured services. Why not just add a `filter` to that to specify what types of accounts you want to retrieve, and then pass `METRICS_STORE` or something similar to it?

It will return anything defined (and not hidden) on the credentials classes, so you would get your `locations` and `recommendedLocations` for free, no?

Those are the 2 new bits we're trying to retrieve, the name & type are already being retrieved.

skandragon: I started adding this to the credentials class, but I think the location is specific to a metric source, not to the generic credential. It didn't feel right to add this to the credentials, so went with the more specific metric service.

kayenta-core/src/main/java/com/netflix/kayenta/security/MetricAccountCredentials.java
skandragon: Ooops.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsService.java
skandragon: Will add.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsService.java
skandragon: Will do.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourceDetail.java
skandragon: Ooops.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourcesController.java
skandragon: Will update.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/security/AtlasNamedAccountCredentials.java
skandragon: Oops.  It's likely just an editor issue...  I notice sometimes if I save a file while the cursor is on a line with blanks, it will preserve them.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
skandragon: Will change this.  It is confusing, and I admit I disliked it as I wrote it.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
skandragon: Will add.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
skandragon: I'm starting to think the UI should do the fallback, and I should just return the literal lists.  This would make our code easier, and the UI can tell the difference should it care.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSourcesController.java
duftler: That's a fair point. Just thought it would be simpler without adding the new service which just ends up back at the credentials anyway.

I can live with it either way.

kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
duftler: Seems reasonable to me.

skandragon: I modified this to return the empty list if there are no recommendations, and let them run independently, with the UI sorting it out.

PR #386, opened 2018-09-27
feat(atlas): Map locations forward to a specific back-end

skandragon:
A previous change was able to generate a list of potential location values to be used by back-ends.  This change uses that in the Atlas code base to look up the specific URI to use for that location string.

This will fall-back to previous behavior if it is not set, and will look in the Atlas-specific context for "dataset", "deployment", and "environment" and use location as the region.

I do not like the big if/else block in Backend.json, but I could not think of a more elegant solution.  I have a test case for all four possibilities, including a no-match for each.

### DIFF
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
@@ -35,6 +35,7 @@ import lombok.*;
 
 import javax.validation.constraints.NotNull;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/model/Backend.java
```diff
@@ -115,4 +116,41 @@ public String getUri(String scheme, String deployment, String dataset, String re
 
     return scheme + "://" + relativeReference;
   }
+
+  public String getUriForLocation(String scheme, String location) {
+    String base = target.replace("$(deployment)", deployment).replace("$(dataset)", dataset);
+    boolean hasRegions = (regions != null && regions.size() > 0);
```
duftler: Could use `!org.springframework.util.CollectionUtils.isEmpty(regions)` && `!CollectionUtils.isEmpty(environments)`

PR #389, opened 2018-10-03
chore(builds): Reformat spinnaker deps version

ttomsu:
The deps as 'XXX' seems a bit troubling... How does this currently work, @duftler ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,10 @@ allprojects {
   apply plugin: "java"
 
   ext {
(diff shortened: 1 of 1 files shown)


ttomsu:
@duftler - Can you grant our team write access to this repo as well?

duftler:
Will have to ask @skandragon why this differs from the other Spinnaker services w.r.t. specifying the spinnaker deps version (note that it appears to be set here: https://github.com/spinnaker/kayenta/blob/master/gradle.properties#L3)

duftler:
Looks like this PR shouldn't be merged until the discrepancy is resolved.

ttomsu:
Ping

ttomsu:
I'm going to push ahead with this change. If it breaks anything, please let me know.

PR #407, opened 2018-10-25
chore(builds): Introduces composite builds via gradle plugin

ttomsu:
Removes spring boot gradle plugin. 

Ref: https://github.com/spinnaker/spinnaker/issues/2844

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,15 +22,14 @@ buildscript {
     maven { url "https://plugins.gradle.org/m2/" }
   }
   dependencies {
-    classpath "com.netflix.spinnaker.gradle:spinnaker-gradle-project:4.0.0"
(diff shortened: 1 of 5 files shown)


fieldju:
The SignalFx integration tests still pass.

I noticed this change eliminates the ability to create a fatJar via `./gradlew bootRepackage -DspringBoot.repackage=true` is there an alternative way to create the fatJar?

duftler:
I don't know the answer to that question. Maybe @ttomsu or @cfieber know.

cfieber:
@fieldju the gradle upgrade includes removal of the spring-boot plugin. If you do want to package a fatjar you could look at creating an init-script to include (similar to what we are doing when publishing is enabled via `-I gradle/init-publish.gradle`)

fieldju:
Makes perfect sense, the init script should do the trick, thanks.

PR #412, opened 2018-10-26
feat(newrelic): implemention for newrelic insights as a metrics store

ehmke:
This is an example PR for issue spinnaker/spinnaker#3464

### DIFF
--- a/kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/metrics/NewrelicMetricsService.java
+++ b/kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/metrics/NewrelicMetricsService.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2018 Adobe
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.newrelic.metrics;
+
+import com.netflix.kayenta.canary.CanaryConfig;
+import com.netflix.kayenta.canary.CanaryMetricConfig;
+import com.netflix.kayenta.canary.CanaryScope;
+import com.netflix.kayenta.canary.providers.metrics.NewrelicCanaryMetricSetQueryConfig;
+import com.netflix.kayenta.metrics.MetricSet;
+import com.netflix.kayenta.metrics.MetricsService;
+import com.netflix.kayenta.newrelic.security.NewrelicCredentials;
+import com.netflix.kayenta.newrelic.security.NewrelicNamedAccountCredentials;
+import com.netflix.kayenta.newrelic.service.NewrelicRemoteService;
+import com.netflix.kayenta.newrelic.service.NewrelicTimeSeries;
+import com.netflix.kayenta.security.AccountCredentialsRepository;
+import com.netflix.spectator.api.Registry;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.validation.constraints.NotNull;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Singular;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.util.StringUtils;
+
+@Builder
+@Slf4j
+public class NewrelicMetricsService implements MetricsService {
+
+  @NotNull
+  @Singular
+  @Getter
+  private List<String> accountNames;
+
+  @Autowired
+  private final AccountCredentialsRepository accountCredentialsRepository;
+
+  @Autowired
+  private final Registry registry;
+
+  @Override
+  public String getType() {
+    return "newrelic";
+  }
+
+  @Override
+  public boolean servicesAccount(String accountName) {
+    return accountNames.contains(accountName);
+  }
+
+  @Override
+  public List<MetricSet> queryMetrics(String accountName,
+      CanaryConfig canaryConfig,
+      CanaryMetricConfig canaryMetricConfig, CanaryScope canaryScope)
+      throws IOException {
+    NewrelicNamedAccountCredentials accountCredentials =
+        (NewrelicNamedAccountCredentials) accountCredentialsRepository
+            .getOne(accountName)
+            .orElseThrow(() -> new IllegalArgumentException(
+                "Unable to resolve account " + accountName + "."));
+
+    NewrelicCredentials credentials = accountCredentials.getCredentials();
+    NewrelicRemoteService remoteService = accountCredentials.getNewrelicRemoteService();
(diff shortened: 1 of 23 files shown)


fieldju:
I am going to UAT this, we use New Relic as well, I want to UAT it by making an integration test similar to [The SignalFx integration test](https://github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/src/integration-test/java/com/netflix/kayenta/signalfx/EndToEndIntegrationTests.java).

kayenta-web/config/kayenta.yml
```diff
@@ -50,6 +50,16 @@ kayenta:
...
 
+  newrelic:
+    enabled: false
+#    accounts:
+#    - name: my-newrelic-account
+#      apiKey: xxxx
+#      applicationKey: xxxx
+#      supportedTypes:
+#      - METRICS_STORE
+#      endpoint.baseUrl: https://insights-api.newrelic.com
```
fieldju: If this endpoint is static and the same for all New Relic customers, then you might consider making it a constant rather then a configuration property.

settings.gradle
```diff
@@ -20,6 +20,7 @@ include 'kayenta-atlas'
 include 'kayenta-aws'
 include 'kayenta-core'
 include 'kayenta-datadog'
+include 'kayenta-newrelic-insights'
```
fieldju: Code style: lexicographical order

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/metrics/NewrelicMetricsService.java
```diff
@@ -0,0 +1,126 @@
...
+    query.append(" SINCE ");
+    query.append(canaryScope.getStart().getEpochSecond());
+    query.append(" UNTIL ");
+    query.append(canaryScope.getEnd().getEpochSecond());
+    query.append(" WHERE ");
+    if (!StringUtils.isEmpty(queryConfig.getQ())) {
+      query.append(queryConfig.getQ());
+      query.append(" AND ");
+    }
+    query.append(canaryScope.getScope());
```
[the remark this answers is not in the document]
fieldju: Here is a visual example of what I am talking about.

image

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/service/NewrelicTimeSeries.java
```diff
@@ -0,0 +1,67 @@
...
+  @Data
+  static public class NewrelicSeriesEntry {
+
+    private Long beginTimeSeconds;
+    private Long inspectedCount;
+    private Long endTimeSeconds;
+    private List<HashMap<String, Number>> results;
+
+    @JsonIgnore
+    private Double adjustSingleResult(HashMap<String, Number> entry) {
```
fieldju: Does New Relic return sparse results, or is it guaranteed to return a complete set of data with missing data set to some sort of empty value?

Exmple of sparse result for some gauge:

```json
{
	"1234": 1,
	"1236": 1,
	"1237": 1
}
```

Notice that that timestamp ```1235``` is missing

example of complete result:

```json
{
	"1234": 1,
	"1235": null,
	"1236": 1,
	"1237": 1
}
```

In either example if the step (data resolution) was set to `1` the following data would be expected.

```json
[1,NaN,1,1]
```

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/metrics/NewrelicMetricsService.java
```diff
@@ -0,0 +1,126 @@
...
+    );
+
+    Instant begin =
+        Instant.ofEpochMilli(timeSeries.getMetadata().getBeginTimeMillis());
+    Instant end =
+        Instant.ofEpochMilli(timeSeries.getMetadata().getEndTimeMillis());
+
+    return Arrays.asList(
+        MetricSet.builder()
+            .name(canaryMetricConfig.getName())
```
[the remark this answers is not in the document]
fieldju: image

fieldju:
@ehmke, I am super excited you opened this PR, this was on my road map for the future, we were planning on just supporting SignalFx to start, but now the New Relic teams will be happy!

I made a couple comments about some gotcha's that aren't super clear, when making a new metrics source. I just finished the SignalFx one, so everything is still fresh.

It would be really nice if you could make an integration test like [this one](https://github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/src/integration-test/java/com/netflix/kayenta/signalfx/EndToEndIntegrationTests.java) that could be used to prove functionality and re-test things when future refactors occur.

Also is your team planning on doing additional PRs to [deck-kayenta](https://github.com/spinnaker/deck-kayenta/pull/351) and [halyard](https://github.com/spinnaker/halyard/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+signalfx) and the [docs](https://github.com/spinnaker/spinnaker.github.io/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+signalfx)?

ehmke:
@fieldju thanks a lot for all that feedback. I am still going through it and reworking my stuff accordingly. I splitted up my changes in several commits and I can squash them later, but maybe it is easier to review, when you have them separate for now.
I am particularly interested in feedback to the third commit "adjusted canaryScope..." where I was trying to fix the canaryScope: as I understood, Spinnaker directly specifies a stack id. Unfortunately, NewRelic doesn't support that, yet. We are still running a custom agent which allows us to specify a git hash - provided during startup time. That is also the reason why we need to narrow down the scope by using the appName field.
I will add some tests and documentation. However, as we are not using halyard, deck or any of the Spinnaker UIs for the type-ahead (we are running kayenta standalone for some reasons), I cannot provide anything there.

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/canary/NewRelicCanaryScope.java
```diff
@@ -26,4 +27,8 @@
 @ToString(callSuper = true)
 public class NewRelicCanaryScope extends CanaryScope {
 
+  @NotNull
+  private String scopeKey;
+
+  private String appName;
```
fieldju: Is this required for all New Relic integrations with Insights or specific to your use case?

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/service/NewrelicTimeSeries.java
fieldju: - [ ] Sanity check that New Relic doesn't return sparse data.

fieldju:
@ehmke sounds good, we can maybe pick up the deck and halyard work.

I have a couple outstanding questions / things to address on this PR. Going to make a check list so I don't lose track.

- [ ] Sanity check that the New Relic API doesn't return sparse results.
- [ ] Address the lack of `step` usage in the New Relic Metrics Service, Does New Relic support passing in this variable, if not it at least be calculated from the returned results.
- [ ] Confirm that appName is globally needed by teams that integrate against New Relic Insights. (If not a generic map of additional query key value pairs might be better / more extensible).

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/service/NewrelicTimeSeries.java
ehmke: New Relic returns complete results. null-values are converted to NaN by our extension.

fieldju:
@ehmke LGTM, I did not user acceptance test any of this, so I am assuming that you tested things after your newest changes.

I think its ready for @duftler and @skandragon to give it their review and merge.

PR #423, opened 2018-11-13
Add more retries to more components

skandragon:
While I would like a more general approach, and this is not quite as clean as I'd like, it's a step toward getting things closer to sane.

I don't yet have metrics wired up, but will add this eventually; I may also split out the "call this method until it stops throwing exceptions" type of call now into a more sane context (http, atlas) aware retry system.  For now, this will work better than what was there.

### DIFF
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
@@ -57,6 +57,9 @@ public class AtlasMetricsService implements MetricsService {
 
   public final String URI_SCHEME = "http";
 
+  public final int MAX_RETRIES = 10; // maximum number of times we'll retry an Atlas query
+  public final long RETRY_BACKOFF = 1000; // time between retries in millis
+
   @NotNull
   @Singular
   @Getter
@@ -151,7 +154,7 @@ public class AtlasMetricsService implements MetricsService {
       atlasResultsList = retry.retry(() -> atlasRemoteService.fetch(decoratedQuery,
                                                                     atlasCanaryScope.getStart().toEpochMilli(),
                                                                     atlasCanaryScope.getEnd().toEpochMilli(), isoStep, credentials.getFetchId(), UUID.randomUUID() + ""),
-      20, 1000);
+      MAX_RETRIES, RETRY_BACKOFF);
     } finally {
       long end = registry.clock().monotonicTime();
       registry.timer("atlas.fetchTime").record(end - start, TimeUnit.NANOSECONDS);
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
@@ -33,6 +33,7 @@ import com.netflix.kayenta.metrics.MetricsService;
 import com.netflix.kayenta.retrofit.config.RemoteService;
 import com.netflix.kayenta.retrofit.config.RetrofitClientFactory;
 import com.netflix.kayenta.security.AccountCredentialsRepository;
+import com.netflix.kayenta.util.Retry;
 import com.netflix.spectator.api.Registry;
 import com.squareup.okhttp.OkHttpClient;
 import lombok.Builder;
@@ -73,6 +74,8 @@ public class AtlasMetricsService implements MetricsService {
   @Autowired
   private final Registry registry;
 
+  private final Retry retry = new Retry();
+
   @Override
   public String getType() {
     return "atlas";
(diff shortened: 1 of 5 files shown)


kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
```diff
@@ -146,12 +148,10 @@ private AtlasNamedAccountCredentials getCredentials(String accountName) {
...
-      atlasResultsList = atlasRemoteService.fetch(decoratedQuery,
-                                                  atlasCanaryScope.getStart().toEpochMilli(),
-                                                  atlasCanaryScope.getEnd().toEpochMilli(),
-                                                  isoStep,
-                                                  credentials.getFetchId(),
-                                                  UUID.randomUUID() + "");
+      atlasResultsList = retry.retry(() -> atlasRemoteService.fetch(decoratedQuery,
+                                                                    atlasCanaryScope.getStart().toEpochMilli(),
+                                                                    atlasCanaryScope.getEnd().toEpochMilli(), isoStep, credentials.getFetchId(), UUID.randomUUID() + ""),
+      20, 1000);
```
duftler: The indentation looks a bit weird here. retry() is passed 3 arguments, right?

duftler: A future reader would have a hard time knowing where the `20` and `1000` came from. Maybe we could use constants for these and `TimeUnit` where it makes sense.

kayenta-core/src/test/groovy/com/netflix/kayenta/util/RetrySpec.groovy
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.util
+
+import spock.lang.Specification
+import spock.lang.Unroll;
+
+class RetrySpec extends Specification {
```
duftler: This file ended up with 4-space indentation instead of 2-space.

kayenta-core/src/main/java/com/netflix/kayenta/util/Retry.java
```diff
@@ -0,0 +1,84 @@
...
+package com.netflix.kayenta.util;
+
+import java.util.function.Supplier;
+
+public class Retry {
+  public void retry(Runnable fn, int maxRetries, long retryBackoff) {
+    retry(fn, maxRetries, retryBackoff, false);
+  }
+
+  public void  exponential(Runnable fn, int maxRetries, long retryBackoff) {
```
duftler: Extraneous space after `void`.

duftler: Please include the units in `retryBackoff`.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -29,7 +29,7 @@
 import com.netflix.kayenta.security.AccountCredentialsRepository;
 import com.netflix.kayenta.storage.ObjectType;
 import com.netflix.kayenta.storage.StorageService;
-import com.netflix.spinnaker.kork.core.RetrySupport;
+import com.netflix.kayenta.util.Retry;
```
duftler: It occurred to me after I finished reviewing the PR that I had seen this logic before. I search kayenta on github and ended up back at this file. Wouldn't it make more sense to just add the new logic to the existing `RetrySupport` class in kork?

kayenta-core/src/test/groovy/com/netflix/kayenta/util/RetrySpec.groovy
skandragon: Fixed.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
skandragon: This is a temporary step along the path to make things more context aware, so I'd prefer to keep it inside kayenta for now.

kayenta-core/src/main/java/com/netflix/kayenta/util/Retry.java
```diff
@@ -0,0 +1,84 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.util;
+
+import java.util.function.Supplier;
+
+public class Retry {
```
Aloren: Please have a look at [spring-retry](https://github.com/spring-projects/spring-retry).
It might be a good solution in case there is a need for more sophisticated retry policies: like configuring which types of exceptions should be retried, adding retry metrics or providing different backoff policies.

Issue #419, opened 2018-11-07
Implement better API error responses


fieldju:
While running Kayenta as a standalone microservice, I noticed that the error response are returned as such

`Unable to resolve account im-memory.` with mime-type `text/plain;charset=UTF-8`

This makes it difficult to handle errors when using this service see the following example.
```java
    CanaryExecutionResponse canaryExecutionResponse;
    try {
      canaryExecutionResponse = kayentaService
          .initiateCanaryWithConfig(stage.getExecution().getApplication(),
              stage.getExecution().getId(), metricsAccount, storageAccount, request);
    } catch (RetrofitError e) {
      String msg;
      try {
        msg = IOUtils.toString(e.getResponse().getBody().in(), "utf-8");
      } catch (Exception e1) {
        msg = e.getMessage();
      }
      throw new RuntimeException("Failed to initiate canary analysis run with Kayenta: " + msg);
    }

    String canaryPipelineExecutionId = canaryExecutionResponse.getCanaryExecutionId();
```

It would be really nice to use a framework such as [Backstopper (framework-agnostic API error handling)](https://github.com/Nike-Inc/backstopper)

With Backstopper all errors would be mime-type `application/json;charset=UTF-8` and follow the same model.

ex:
```json
{
  "error_id": "408d516f-68d1-41f3-adb1-b3dc0affaaf2",
  "errors": [
    {
      "code": 43,
      "message": "Unable to resolve account im-memory."
    }
  ]
}
```

[EDIT] This also gives a uuid that can be used to find the additional details in the logs.
With this using Kayenta from other microservices would be nicer. 

ex:
```java
    CanaryExecutionResponse canaryExecutionResponse;
    try {
      canaryExecutionResponse = kayentaService
          .initiateCanaryWithConfig(stage.getExecution().getApplication(),
              stage.getExecution().getId(), metricsAccount, storageAccount, request);
    } catch (RetrofitError e) {
      ApiError error = (ApiError) e.getBodyAs(ApiError.class);
      throw new RuntimeException("Failed to initiate canary analysis run with Kayenta: " + 
          String.join(", ", error.getErrors().stream()
              .map(error -> error.getMessage())
              .collect(Collectors.toList())));
    }
```

This would change the mime-type of the error responses, so maybe thats too much of a change?

If this is something that sounds like a good idea please let me know and I can make a PR or else feel free to close this issue.

nicmunroe:
(👋 Primary maintainer of Backstopper here) 

[...]

PR #430, opened 2018-11-21
Fixes content-type in error response for IllegalArgumentException

Aloren:
- error response now is sent in json format

Relates to #419


Aloren:
@skandragon could you please review the changes? Thanks

fieldju:
@duftler,

I am in favor of merging this.
This will make the mime-types of Kayenta responses consistent (application/json).

This is technically a breaking API change though, and I am not sure what the consequences of this change would be to existing integrations.

I still want to do a spike later on Backstopper for Request validation and error handling and see that pattern would be nice and work with this code base.

skandragon:
@duftler Do you see any reason that deck-kayenta would need to be changed for this to be merged?  It seems likely it will just work, and feels like a good change.

duftler:
The changes look good to me, but might be worth running orca against it and confirming that orca can still propagate the returned exception properly. I'm sure deck-kayenta just renders what orca returns, so the question is if orca needs to be changed.

PR #431, opened 2018-11-22
feat(newrelic): migrate towards more flexible queries.

ehmke:
This requires the canary config to contain the first full part of the NRQL select statement and allows to query from other stores (beside "transaction").


fieldju:
I didn't get the original notification to review this for some reason.

PR #433, opened 2018-11-26
feat(prometheus): add customInlineTemplate logic to enable metric-spec…

maggieneterval:
…ific PromQL queries

- Rename `CanaryMetricSetQueryConfig.customFilter` to `customInlineTemplate` to clarify that this template belongs to a single metric within a canary config (vs. `customFilterTemplate`, which is a key that references a template shared across metrics within a canary config). Although technically a breaking change, this field had not previously been exposed in the UI. [This PR](https://github.com/spinnaker/deck-kayenta/pull/387) adds UI support for editing full PromQL queries inline as `customInlineTemplate`.
- Adjust `QueryConfigUtils.expandCustomFilter` to apply the same template expansion logic to `customInlineTemplate` and `customFilterTemplate`. This will allow users to hydrate PromQL query templates with custom and implicitly available parameters.

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryMetricSetQueryConfig.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryMetricSetQueryConfig.java
@@ -24,8 +24,17 @@ import lombok.NonNull;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public interface CanaryMetricSetQueryConfig {
 
-  // Optionally defines an explicit filter to be used when composing the query. Takes precedence over customFilterTemplate.
+  /**
+   * @deprecated Use getCustomInlineTemplate instead.
+   */
+  @Deprecated
   default String getCustomFilter() {
+    return null;
+  }
+
+  // Optionally defines a metric-specific query template. Takes precedence over customFilterTemplate.
+  // It is expanded by using the key/value pairs in extendedScopeParams as the variable bindings.
+  default String getCustomInlineTemplate() {
     return null;
   }
 
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
@@ -49,13 +49,25 @@ public class QueryConfigUtils {
                                           CanaryMetricSetQueryConfig metricSetQuery,
                                           CanaryScope canaryScope,
                                           String[] baseScopeAttributes) throws IOException {
-    String customFilter = metricSetQuery.getCustomFilter();
-    String customFilterTemplate = metricSetQuery.getCustomFilterTemplate();
+    if (metricSetQuery.getCustomFilter() != null) {
+      throw new IllegalArgumentException("CanaryMetricSetQueryConfig.customFilter is deprecated, use CanaryMetricSetQueryConfig.customInlineTemplate instead.");
+    }
 
-    log.debug("customFilter={}", customFilter);
+    String customInlineTemplate = metricSetQuery.getCustomInlineTemplate();
+    String customFilterTemplate = metricSetQuery.getCustomFilterTemplate();
+    String templateToExpand;
+    String expandedTemplate;
+
+    log.debug("customInlineTemplate={}", customInlineTemplate);
     log.debug("customFilterTemplate={}", customFilterTemplate);
 
-    if (StringUtils.isEmpty(customFilter) && !StringUtils.isEmpty(customFilterTemplate)) {
+    if (StringUtils.isEmpty(customInlineTemplate) && StringUtils.isEmpty(customFilterTemplate)) {
+      return null;
+    }
+
+    if (!StringUtils.isEmpty(customInlineTemplate)) {
+      templateToExpand = unescapeTemplate(customInlineTemplate);
+    } else {
       Map<String, String> templates = canaryConfig.getTemplates();
 
       // TODO(duftler): Handle this as a config validation step instead.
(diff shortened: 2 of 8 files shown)


kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
```diff
@@ -49,49 +49,59 @@ public static String expandCustomFilter(CanaryConfig canaryConfig,
...
+      templateToExpand = unescapeTemplate(customInlineTemplate);
+    } else {
       Map<String, String> templates = canaryConfig.getTemplates();
 
       // TODO(duftler): Handle this as a config validation step instead.
       if (CollectionUtils.isEmpty(templates)) {
-        throw new IllegalArgumentException("Custom filter template '" + customFilterTemplate + "' was referenced, " +
-                                           "but no templates were defined.");
+        throw new IllegalArgumentException(
+            "Custom filter template '" + customFilterTemplate + "' was referenced, " +
```
duftler: If we keep the formatting the same for this exception and the one just below, it'll cut down on the number of changed lines in the PR.

kayenta-stackdriver/src/test/groovy/com/netflix/kayenta/providers/metrics/StackdriverCanaryMetricSetQueryConfigSpec.groovy
```diff
@@ -54,23 +54,23 @@ class StackdriverCanaryMetricSetQueryConfigSpec extends Specification {
...
     StackdriverCanaryScope stackdriverCanaryScope = new StackdriverCanaryScope(extendedScopeParams: scopeParams)
 
     expect:
     QueryConfigUtils.expandCustomFilter(canaryConfig, stackdriverCanaryMetricSetQueryConfig, stackdriverCanaryScope) == expectedExpandedTemplate
 
     where:
-    templates                                             | customFilterTemplate | customFilter          | scopeParams       || expectedExpandedTemplate
-    ["my-template": 'A test: key1=${key1}.']              | "my-template"        | "An explicit filter." | [key1: "value-1"] || "An explicit filter."
+    templates                                             | customFilterTemplate | customInlineTemplate          | scopeParams       || expectedExpandedTemplate
+    ["my-template": 'A test: key1=${key1}.']              | "my-template"        | "An inline template."         | [key1: "value-1"] || "An inline template."
```
duftler: Since the `customInlineTemplate` values will now be expanded, should we include a couple of tests that have variables to be bound?

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryMetricSetQueryConfig.java
```diff
@@ -24,8 +24,9 @@
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public interface CanaryMetricSetQueryConfig {
 
-  // Optionally defines an explicit filter to be used when composing the query. Takes precedence over customFilterTemplate.
-  default String getCustomFilter() {
+  // Optionally defines a metric-specific query template. Takes precedence over customFilterTemplate.
+  // It is expanded by using the key/value pairs in extendedScopeParams as the variable bindings.
+  default String getCustomInlineTemplate() {
```
duftler: Should probably run a manual test to ensure that previously-created configs that were persisted with the original `customFilter` attribute don't cause the updated deserialization logic to choke.

duftler: May make sense to allow for configs with the deprecated attribute to be deserialized and then throw an exception so the config owner knows to update the config.

kayenta-stackdriver/src/test/groovy/com/netflix/kayenta/providers/metrics/StackdriverCanaryMetricSetQueryConfigSpec.groovy
```diff
@@ -54,23 +54,23 @@ class StackdriverCanaryMetricSetQueryConfigSpec extends Specification {
...
-     "my-template-2": 'A test: key2=${key2}.']            | "my-template-1"      | "An explicit filter." | [key1: "value-1"] || "An explicit filter."
-    ["my-template": 'A test: key1=${key1} key2=${key2}.'] | "my-template"        | "An explicit filter." | [key1: "value-1",
-                                                                                                            key2: "value-2"] || "An explicit filter."
-    ["my-template": 'A test: key1=something1.']           | "my-template"        | "An explicit filter." | null              || "An explicit filter."
-    ["my-template": 'A test: key1=something1.']           | "my-template"        | "An explicit filter." | [:]               || "An explicit filter."
+     "my-template-2": 'A test: key2=${key2}.']            | "my-template-1"      | "An inline template."          | [key1: "value-1"] || "An inline template."
+    ["my-template": 'A test: key1=${key1} key2=${key2}.'] | "my-template"        | "An inline template."          | [key1: "value-1",
+                                                                                                                    key2: "value-2"]  || "An inline template."
```
duftler: Looks like (at least from the github diff view) that `key1` and `key2` as misaligned here.

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryMetricSetQueryConfig.java
duftler: If you move `customFilter` to the base class, you can throw the exception right in `QueryConfigUtils.expandCustomFilter()`. At least this way it will only notify when someone tries to actually use one of those configs.

maggieneterval: Awesome thanks, updated. Tested to make sure we get the error back in the UI as well:
pyazzjrqbx2

kayenta-stackdriver/src/test/groovy/com/netflix/kayenta/providers/metrics/StackdriverCanaryMetricSetQueryConfigSpec.groovy
maggieneterval: Great catch, impressed that you noticed that in the github diff view! Fixed.

maggieneterval:
@spinnakerbot cherry-pick 1.11

Issue #288, opened 2018-04-11
Support Graphite as a source


aivans:
Many early adopters of the Netflix/Spring Cloud stacks still use Graphite for as their metrics backend system.
It would be useful to offer support for Graphite.

lwander:
If anyone in the community is interested in adding Graphite support they can use the DataDog integration as a template: https://github.com/spinnaker/kayenta/pull/248

spinnakerbot:
This issue hasn't been updated in 49 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

duftler:
Backend support was added here: https://github.com/spinnaker/kayenta/pull/447

duftler:
Frontend support was added here: https://github.com/spinnaker/deck-kayenta/pull/399

PR #447, opened 2018-12-20
feat(graphite): add Graphite integration

ybai2sc:
add Graphite as metric store address https://github.com/spinnaker/kayenta/issues/288


kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/controller/GraphiteFetchController.java
```diff
@@ -0,0 +1,120 @@
...
+                                   SynchronousQueryProcessor synchronousQueryProcessor,
+                                   GraphiteConfigurationTestControllerDefaultProperties graphiteConfigurationTestControllerDefaultProperties) {
+        this.accountCredentialsRepository = accountCredentialsRepository;
+        this.synchronousQueryProcessor = synchronousQueryProcessor;
+        this.graphiteConfigurationTestControllerDefaultProperties =
+                graphiteConfigurationTestControllerDefaultProperties;
+    }
+
+    @RequestMapping(value = "/query", method = RequestMethod.POST)
+    public Map queryMetrics(@RequestParam(required = false) final String metricsAccountName,
```
duftler: Shouldn't `location` also be supported here, instead of passing `null`?

kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/controller/GraphiteFetchController.java
```diff
@@ -0,0 +1,120 @@
...
+                            @RequestParam(required = false) final String storageAccountName,
+                            @ApiParam(defaultValue = "cpu") @RequestParam String metricSetName,
+                            @ApiParam(defaultValue = "system.cpu.user") @RequestParam String metricName,
```
duftler: Seems like the `defaultValue` (or descriptive text) for `metricName` should give an example of using the templating that the metrics service's approach relies upon. That is, it should be clear to the user that they may want to include `$scope` or `$location` directly in `metricName`.

kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/controller/GraphiteFetchController.java
ybai2sc: yes, thanks for catching it

kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/controller/GraphiteFetchController.java
ybai2sc: thx, will add `$location` and `$scope` in default value and add some description in `location` and `scope`

kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/metrics/GraphiteMetricsService.java
```diff
@@ -0,0 +1,202 @@
...
+            GraphiteNamedAccountCredentials accountCredentials =
+                    (GraphiteNamedAccountCredentials) accountCredentialsRepository.getOne(metricsAccountName)
+                            .orElseThrow(() -> new IllegalArgumentException(
+                                    String.format("Unable to resolve account %s.", metricsAccountName)));
+
+            GraphiteRemoteService remoteService = accountCredentials.getGraphiteRemoteService();
+
+            filter = convertFilterToGraphiteQuery(filter);
+
+            GraphiteMetricDescriptorsResponse graphiteMetricDescriptorsResponse =
```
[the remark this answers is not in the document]
ybai2sc: yes, it responds kind of sufficiently well to ui. if there are lots of users, then the response may not be fast enough.
the current problem of caching is that there will not a lots of metric entries in graphite server and when it reaches an amount, the server will not respond the get all entries request.  so we have to use query level by level to get the typeahead working.  for example the query is a.b.c.d.e , when user types a. the service will only return a.b. 

cache the last response is a good idea, could we leave a TODO here in case we hit performance issue ?

kayenta-graphite/README.md
```diff
@@ -0,0 +1,49 @@
...
+          - METRICS_STORE
+```
+
+## Development
+
+### Integration Tests
+
+This module has an End to End test with in memory store, embedded redis and embedded graphite server.
+
+It creates three threads that feeds data to embedded graphite server with different value.
```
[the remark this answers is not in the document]
ybai2sc: thx , will fix them

ybai2sc:
@duftler  thanks for reviewing it!  I used 4 space indentation in this project. some 8 space indentation you saw probably is just for line break when it reaches 120, for example https://github.com/spinnaker/kayenta/pull/447/files#diff-1e0e28962a7681fbc3210b94694f561dR115

duftler:
> @duftler thanks for reviewing it! I used 4 space indentation in this project. some 8 space indentation you saw probably is just for line break when it reaches 120, for example https://github.com/spinnaker/kayenta/pull/447/files#diff-1e0e28962a7681fbc3210b94694f561dR115

Even for line breaks, I believe we're using 4 or 2 everywhere.

kayenta-graphite/src/main/java/com/netflix/kayenta/graphite/metrics/GraphiteMetricsService.java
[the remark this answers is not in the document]
duftler: Yes, sounds good to me.

ybai2sc:
I see, will change those to 4 spaces

PR #448, opened 2018-12-29
chore(stackdriver): Use example instead of defaultValue in anticipati…

duftler:
…on of swagger upgrade.


duftler:
The update causes an exception to be thrown on trying to parse an empty string into a Long.

Issue #456, opened 2019-01-07
Support Basic Auth for metric backends


Remz-Jay:
In our case, specifically for the Prometheus backend.

Our Prometheus installations are protected by AAA to ensure we don't have unauthorized access to our metrics. To allow system accounts to access to Prometheus, we have al alternate Kubernetes Ingress that serves the "prometheus-api" endpoint, which is protected by Basic Auth.

As such, Spinnaker/Kayenta needs to use this endpoint and provide basic auth credentials. I've done so by adding them to the `--base-url` argument in https://github.com/spinnaker/halyard/blob/master/docs/commands.md#hal-config-canary-prometheus-account-add .

After that deploys, Kayenta fails with the following error:

```
2019-01-07 14:37:41.197  INFO 1 --- [pool-1-thread-1] c.n.s.o.t.RedisActiveExecutionsMonitor   : Refreshing active execution gauges (active: 0)
2019-01-07 14:37:53.624  INFO 1 --- [    scheduler-2] c.n.k.p.service.PrometheusRemoteService  : ---> HTTP GET https://spinnaker:password@prometheus-api.some.domain/api/v1/label/__name__/values
2019-01-07 14:37:53.715  INFO 1 --- [    scheduler-2] c.n.k.p.service.PrometheusRemoteService  : <--- HTTP 401 https://spinnaker:password@prometheus-api.some.domain/api/v1/label/__name__/values (90ms)
2019-01-07 14:37:53.717 ERROR 1 --- [    scheduler-2] o.s.s.s.TaskUtils$LoggingErrorHandler    : Unexpected error occurred in scheduled task.

retrofit.RetrofitError: 401 Unauthorized
  at retrofit.RetrofitError.httpError(RetrofitError.java:40) ~[retrofit-1.9.0.jar:na]
  at retrofit.RestAdapter$RestHandler.invokeRequest(RestAdapter.java:388) ~[retrofit-1.9.0.jar:na]
  at retrofit.RestAdapter$RestHandler.invoke(RestAdapter.java:240) ~[retrofit-1.9.0.jar:na]
  at com.sun.proxy.$Proxy96.listMetricDescriptors(Unknown Source) ~[na:na]
  at com.netflix.kayenta.prometheus.metrics.PrometheusMetricsService.updateMetricDescriptorsCache(PrometheusMetricsService.java:337) ~[kayenta-prometheus.jar:na]
[...]

PR #460, opened 2019-01-10
feat(prometheus): Add support for basic auth attribute on Prometheus …

duftler:
…accounts.


duftler:
Add support for basic auth. Addresses #456.

PR #462, opened 2019-01-18
fix: When AWS and S3 are enabled properties of type primitive boolean…

fieldju:
@duftler @skandragon ,

I noticed that when I have Kayenta running with AWS and S3 configured that response objects with properties that are of type boolean are being stripped from the responses.

This PR changes a couple booleans to Boolean.

It might be better long term to figure out what is going on with Jackson?


fieldju:
We can decline this PR if the solution in https://github.com/spinnaker/kayenta/issues/464 sounds good, as this PR is more a bandaid that a real fix.

PR #470, opened 2019-01-23
feat(signalfx): add optional default scope and location keys when configuring accounts

fieldju:
This PR makes including `_scope_key` in the extendedScopeParameters optional and allows operators to define account / org wide defaults.

This PR also adds the location to the query if it is supplied as an account default or in the extendedScopeParameters as `_location_key`.

Both changes are backwards compatible.

### DIFF
--- a/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
+++ b/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
@@ -1,0 +1,17 @@
+package com.netflix.kayenta.signalfx.config;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class SignalFxScopeConfiguration {
+
+  private String defaultScopeKey;
+  private String defaultLocationKey;
+
+}
--- a/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
+++ b/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
@@ -13,5 +13,4 @@ public class SignalFxScopeConfiguration {
 
   private String defaultScopeKey;
   private String defaultLocationKey;
-
 }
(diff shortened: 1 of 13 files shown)


kayenta-signalfx/src/test/java/com/netflix/kayenta/signalfx/metrics/SimpleSignalFlowProgramBuilderTest.java
```diff
@@ -58,13 +60,88 @@ public void test_that_the_program_builder_builds_the_expected_program() {
     assertEquals(expected, builder.build());
   }
 
+  @Test
+  public void test_that_the_program_builder_builds_the_expected_program_when_location_is_provided_via_default() {
```
duftler: This test and the one just below seem to match except for where the location key is specified. Could combine them and use a `where:` block and `@Unroll` instead. Would also make it easier for a reader to tease out the differences.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
```diff
@@ -0,0 +1,17 @@
...
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class SignalFxScopeConfiguration {
+
+  private String defaultScopeKey;
+  private String defaultLocationKey;
+
```
duftler: Extraneous blank line.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
```diff
@@ -80,13 +84,15 @@ public String buildQuery(String metricsAccountName,
                            CanaryMetricConfig canaryMetricConfig,
                            CanaryScope canaryScope) {
 
-    SignalFxCanaryScope signalFxCanaryScope = (SignalFxCanaryScope)canaryScope;
+    SignalFxScopeConfiguration scopeConfiguration = signalFxScopeConfigurationMap.get(metricsAccountName);
+
```
duftler: Extraneous blank line.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/config/SignalFxScopeConfiguration.java
fieldju: - [X] done.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
fieldju: - [X] done.

PR #486, opened 2019-02-13
feat(canary-results): archive exactly what we return via HTTP

skandragon:
This change allows execution details to be removed from the Redis-backed execution repository, yet still return JSON data from object storage.  This makes the redis footprint smaller while still maintaining long-lived links to canary reports or raw data.

Use object storage to archive the same response we'd give back for `GET /canary/:id`.  Add a controller to manipulate these archive objects.  The controller is documented as not intended to be used directly.

In the CanaryController, first look in redis for the execution information, and if found return it as usual.  If not found, look in the archive and return what we have from that.

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
@@ -16,14 +16,12 @@
 package com.netflix.kayenta.canary;
 
 import com.netflix.kayenta.canary.results.CanaryResult;
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.*;
 
 import javax.validation.constraints.NotNull;
 import java.util.Map;
 
+@Getter
 @Data
 @Builder
 @NoArgsConstructor
@@ -34,7 +32,7 @@ public class CanaryExecutionStatusResponse {
   protected String parentPipelineExecutionId;
 
   @NotNull
-  protected String pipelineId;
+  public String pipelineId;
 
   @NotNull
   protected Map<String, String> stageStatus;
@@ -66,14 +64,22 @@ public class CanaryExecutionStatusResponse {
   // (endTime - buildTime) should indicate the total time it took from request to result.
   // (endTime - startTime) should be the amount of time the canary was actually running.
   //
+
   protected Long buildTimeMillis;
+
   protected String buildTimeIso;
+
   protected Long startTimeMillis;
+
   protected String startTimeIso;
+
   protected Long endTimeMillis;
+
   protected String endTimeIso;
 
   // If set, these are the account names used for this run.
+
   protected String storageAccountName;
+
   protected String configurationAccountName;
 }
--- a/kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java
@@ -18,6 +18,7 @@ package com.netflix.kayenta.storage;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.netflix.kayenta.canary.CanaryConfig;
+import com.netflix.kayenta.canary.CanaryExecutionStatusResponse;
 import com.netflix.kayenta.canary.results.CanaryResult;
 import com.netflix.kayenta.metrics.MetricSet;
 import com.netflix.kayenta.metrics.MetricSetPair;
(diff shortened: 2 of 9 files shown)


skandragon:
@duftler do you see any reason in deck-kayenta that we should not merge this?

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
```diff
@@ -36,6 +36,11 @@
   @NotNull
   protected String pipelineId;
 
+  @NotNull
+  public String getPipelineId() {
```
duftler: Doesn't this come for free from `@Data`?

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
```diff
@@ -74,6 +80,12 @@
   protected String endTimeIso;
 
   // If set, these are the account names used for this run.
+
   protected String storageAccountName;
+
+  public String getStorageAccountName() {
```
duftler: Same question as up above re lombok. I thought `@Data` meant `@Getter` is implicitly applied. Maybe it doesn't work on `protected` fields? If so, can just add `@Getter` to the needed fields, no?

kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java
```diff
@@ -31,7 +32,9 @@
   CANARY_CONFIG(new TypeReference<CanaryConfig>() {}, "canary_config", "canary_config.json"),
   METRIC_SET_LIST(new TypeReference<List<MetricSet>>() {}, "metrics", "metric_sets.json"),
   METRIC_SET_PAIR_LIST(new TypeReference<List<MetricSetPair>>() {}, "metric_pairs", "metric_set_pairs.json"),
-  CANARY_RESULT(new TypeReference<CanaryResult>() {}, "canary_results", "canary_results.json");
+  CANARY_RESULT_ARCHIVE(new TypeReference<CanaryExecutionStatusResponse>() {}, "canary_archive", "canary_archive.json"),
+  CANARY_RESULT(new TypeReference<CanaryResult>() {}, "judge_results", "judge_results.json");
+
```
duftler: Extraneous blank line.

kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java
```diff
@@ -31,7 +32,9 @@
   CANARY_CONFIG(new TypeReference<CanaryConfig>() {}, "canary_config", "canary_config.json"),
   METRIC_SET_LIST(new TypeReference<List<MetricSet>>() {}, "metrics", "metric_sets.json"),
   METRIC_SET_PAIR_LIST(new TypeReference<List<MetricSetPair>>() {}, "metric_pairs", "metric_set_pairs.json"),
-  CANARY_RESULT(new TypeReference<CanaryResult>() {}, "canary_results", "canary_results.json");
+  CANARY_RESULT_ARCHIVE(new TypeReference<CanaryExecutionStatusResponse>() {}, "canary_archive", "canary_archive.json"),
```
duftler: We don't have a proper sort order here, but I _think_ I would have put `CANARY_RESULT_ARCHIVE` after `CANARY_RESULT` for reasons I can't really explain.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryResultArchiveController.java
```diff
@@ -0,0 +1,142 @@
...
+
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+@RestController
+@RequestMapping("/canaryResultArchive")
+@Api(value = "/canaryResultArchive", description = "Manipulate the arvhived canary result object store.  This should be used only for Kayenta maintenance.  Use the /canary endpoints for canary results.")
```
duftler: s/arvhived/archived

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
skandragon: You'd think so, but there is something very strange happening about protection.  I used `@Getter` which did not allow access, and ended up rolling my own getter for this use case.

kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java
skandragon: Heh.  OK.  :)

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryResultArchiveController.java
skandragon: You have no idea how many times I misspelled that word...

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryExecutionStatusResponse.java
skandragon: Ah ha, adding `@Getter` as a default for the class works, even though adding it to individual fields does not.

PR #496, opened 2019-02-20
fix(orca): Update Orca to 6.139.0

dibyom:
* chore(orca): Bump Orca to 6.139.0
* fix(canary/orca): Replace limit with pageSize 
The `limit` parameter was replaced with `pageSize` in
spinnaker/orca#2593


duftler:
@skandragon Any concerns?

dibyom:
@spinnakerbot cherry-pick 1.12

PR #526, opened 2019-04-17
feat(atlas): Find atlas back-end by accountId

skandragon:
Add support for going from an accountId (usually an AWS account ID string) to an atlas backend which stores data for that account.

This change adds another magic URL (which I should document) to the Atlas back-end.  This database consists of a set of account IDs to Atlas back-ends.  This makes kayenta-atlas more Netflix specific, but I can address this later should anyone show interest in using Atlas outside of Netflix.

### DIFF
--- a/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
+++ b/kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
@@ -16,11 +16,13 @@
 
 package com.netflix.kayenta.atlas.metrics;
 
+import com.netflix.kayenta.atlas.backends.AtlasStorageDatabase;
 import com.netflix.kayenta.atlas.backends.BackendDatabase;
 import com.netflix.kayenta.atlas.canary.AtlasCanaryScope;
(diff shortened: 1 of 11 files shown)


kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/metrics/AtlasMetricsService.java
```diff
@@ -115,6 +117,34 @@ private AtlasNamedAccountCredentials getCredentials(String accountName) {
     AtlasNamedAccountCredentials credentials = getCredentials(accountName);
     BackendDatabase backendDatabase = credentials.getBackendUpdater().getBackendDatabase();
     String uri = backendDatabase.getUriForLocation(URI_SCHEME, atlasCanaryScope.getLocation());
+
```
fieldju: Its a bit of a nit, but the URI resolving logic is starting to get a bit long, maybe it would make sense to refactor / extract it into its own method at some point.

If you were running something like Sonar it might complain about the `Cyclomatic complexity` of this method.

skandragon: Yea, I totally understand.  This is all going to be cleaned up -- soon.  For now, though, I'll merge.

PR #534, opened 2019-05-17
fix(datadog): Create MetricSet placeholder

gregorymfoster:
**fix(datadog): Support placeholder metrics in `DatadogMetricsService`**

DatadogMetricsService currently returns an empty `List<MetricSet>` if no metrics can be found for some `canaryConfig` and `scope`.  This can result in complete CanaryAnalysis stage failure if one metric cannot be found for a CanaryConfig.

This can be frustrating in several cases:
- An application chooses to stop publishing a metric, but does not update the CanaryConfig.
- An application publishes some, but not all, metrics needed for a shared CanaryConfig that it references.

Kayenta already gracefully supports a lack of MetricSet values. This change will leverage that support by creating an empty placeholder metric when Datadog fails to return any metric.

References:
[Stackdriver placeholder example](https://github.com/spinnaker/kayenta/blob/master/kayenta-stackdriver/src/main/java/com/netflix/kayenta/stackdriver/metrics/StackdriverMetricsService.java#L316-L319)
[Error raised without Datadog placeholder](https://github.com/spinnaker/kayenta/blob/084a1a7d771ba1d42a54510bf730c796dbb891a5/kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricSetMixerService.java#L184)

UI failure when an existing metric cannot be found for a control scope tag

### DIFF
--- a/kayenta-datadog/src/main/java/com/netflix/kayenta/datadog/metrics/DatadogMetricsService.java
+++ b/kayenta-datadog/src/main/java/com/netflix/kayenta/datadog/metrics/DatadogMetricsService.java
@@ -133,6 +133,22 @@ public class DatadogMetricsService implements MetricsService {
       );
     }
 
+    if (ret.isEmpty()) {
+      // Add placeholder metric set.
+      ret.add(
+        MetricSet.builder()
+          .name(canaryMetricConfig.getName())
+          .startTimeMillis(canaryScope.getStart().toEpochMilli())
+          .startTimeIso(canaryScope.getStart().toString())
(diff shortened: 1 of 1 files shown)


fieldju:
LGTM @duftler is the gate keeper on this one these days.

gregorymfoster:
@fieldju thank you for taking a look! :)

gregorymfoster:
cc @benjaminws

cfieber:
@duftler there was a question in slack from @gregorymfoster on whether this could get cherry-picked to 1.14

duftler:
@spinnakerbot cherry-pick 1.14

PR #565, opened 2019-06-19
feat(kayenta-bom): publish kayenta-bom

marchello2000:
This makes downstream dependencies easier.
NOTE: Exposing `orca-bom` which then brings in `kork-bom`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,37 +37,40 @@ allprojects {
   if (Boolean.valueOf(enablePublishing)) {
     apply plugin: "spinnaker.project"
   }
-  apply plugin: "java-library"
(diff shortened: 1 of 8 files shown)


cfieber:
opened this so we don't write korkVersion back into gradle.properties:
https://github.com/spinnaker/spinnaker-gradle-project/pull/75

marchello2000:
> opened this so we don't write korkVersion back into gradle.properties:
> [spinnaker/spinnaker-gradle-project#75](https://github.com/spinnaker/spinnaker-gradle-project/pull/75)

ah, sorry, i forgot to do that. thanks!

PR #568, opened 2019-06-21
Bad requests to metrics storage are retried, but should not

### LINKED ISSUE #436: Bad requests to metrics storage are retried, but should not
Mechanism of retries that was introduced in #423 retries requests that return BAD REQUEST status. I've checked that for influxdb implemenetation and that seems strange to me. 
See:
```
      } catch (IOException | UncheckedIOException | RetrofitError | RetryableQueryException e) {
        retries++;
        // TODO: Externalize this as a configurable setting.
        if (retries >= 10)
          throw e;
        log.warn("Retrying metric service query");
      }
```
I think that the better way of retrying retrofit errors is by using custom interceptor where retry policy can be configured more correctly.

Aloren:
Bad requests to metrics storage are retried, but should not 
fixes #436


kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
```diff
@@ -96,6 +112,12 @@ public String executeQuery(String metricsAccountName,
     return metricSetListId;
   }
 
+  private boolean isRetryable(RetrofitError e) {
```
fieldju: Am I to understand that the new default logic is to only retry 5xx response, where as the old logic was to retry any non 200-399 responses?

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
```diff
@@ -0,0 +1,17 @@
...
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.http.HttpStatus;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Data
+@ConfigurationProperties("kayenta.retry")
```
fieldju: Can you add an example of configuring this to https://github.com/spinnaker/kayenta/blob/master/kayenta-web/config/kayenta.yml

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
Aloren: Yes, you are correct. Should I add other http series into default configuration as well? The intent of this PR was to exclude 4xx response statuses from retries.

fieldju: I don't know if there is a real use case for retrying other responses (0-199, 300+) or not.

**I was just making sure I understand what is changing.**

The situation that comes to my mind that would warrant a retry are 503s (LB issues) and timeouts (network issues)

fieldju: image

What about 409s, I don't think I have actually ever seen that response code in real life, but that seem retry-able.

image

I have seen 429's and they are retry-able usually with some sort of back off policy

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
```diff
@@ -0,0 +1,17 @@
...
+public class MetricsRetryConfigurationProperties {
+
+    private List<HttpStatus.Series> series = Arrays.asList(HttpStatus.Series.SERVER_ERROR);
```
fieldju: Is it worth adding 408s and 429s to this list and make people opt out of those codes rather than opt in, to maintain backwards compatibility.

Aloren: Ok, I will also add http statuses list that can be retried

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
Aloren: There is no any backoff in the current retry mechanism, so I'm not sure that it makes sense to retry 429 status by default. 
In any case I've already added both `REQUEST_TIMEOUT` and `TOO_MANY_REQUESTS` to default configuration.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
Aloren: yes, of course. I've added default configuration as an example into `kayenta.yml`.
What do you think regarding naming? `kayenta.retry` is ok prefix of better to rename to `kayenta.metrics.retry`?

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
```diff
@@ -81,10 +87,20 @@ public String executeQuery(String metricsAccountName,
...
         success = true;
-      } catch (IOException | UncheckedIOException | RetrofitError | RetryableQueryException e) {
+      } catch (RetrofitError e) {
+
+        boolean retryable = isRetryable(e);
+        if (retryable) {
+          retries++;
+          if (retries >= retryConfiguration.getAttempts())
+            throw e;
+          log.warn("Retrying metric service query");
```
duftler: This log message won't give the user much to go on when troubleshooting. Can you include some details that will help when tracing requests?

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
```diff
@@ -0,0 +1,28 @@
+package com.netflix.kayenta.metrics;
```
duftler: Please add a copyright header to this file.

kayenta-web/config/kayenta.yml
```diff
@@ -115,6 +115,12 @@ kayenta:
 #    endpoint:
 #      baseUrl: http://remotejudge.example.com/path/to/service
 
+# Retry configuration for metrics fetching from metrics storage (see: MetricsRetryConfigurationProperties)
+  retry:
```
duftler: Would prefer `kayenta.metrics.retry`, as you mentioned.

duftler:
If this PR does change the set of status codes that are retried, we will need someone from Netflix to verify the behavior against Atlas.
@erikmunson

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
Aloren: fixed error message for both cases

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
Aloren: added

kayenta-web/config/kayenta.yml
Aloren: fixed

kayenta-core/src/test/groovy/com/netflix/kayenta/metrics/SynchronousQueryProcessorTest.java
```diff
@@ -0,0 +1,147 @@
+package com.netflix.kayenta.metrics;
```
duftler: Missing copyright header.

kayenta-core/src/main/java/com/netflix/kayenta/metrics/MetricsRetryConfigurationProperties.java
duftler: Thanks. Please use 2019 instead of 2017 and your name or company name instead of Google.

Aloren:
rebased on master

PR #570, opened 2019-06-25
Workaround prom templating issues

jmccann:
This PR is a workaround issues we've had using templating with prometheus.  For more context see https://github.com/spinnaker/spinnaker/issues/4235

Sorry ahead of time for my use of terminology.  I'm a spinnaker n00bie and may misuse terms.

It seems that the prometheus service is properly able to parse the template and fetch the appropriate metrics when watching the logs.  However, orca? `c.n.s.o.p.u.ContextParameterProcessor` is also trying to process the templating and fails which fails the whole pipeline.

This PR works around that by not passing any `${}` anywhere so orca doesn't try to process it and fail.  We added some query config parameters to still allow building "complex" prom queries without using `${}` templating.

Basically, with a metric query config like:

```
        "query": {
          "type": "prometheus",
          "metricName": "num_of_404s",
          "customInlineTemplate": "PromQL:sum(irate(statuses{code=\"404\",kubernetes_name=\"${scope}\"}[5m]))",
          "serviceType": "prometheus"
        },
...
    "scopes": [
      {
        "scopeName": "default",
        "controlScope": "my-control",
        "experimentScope": "my-experiment",
        "step": 10
      }
    ],
```

We would expect queries of:
* control: `PromQL:sum(irate(statuses{code="404",kubernetes_name="my-control"}[5m]))`
* experiment: `PromQL:sum(irate(statuses{code="404",kubernetes_name="my-experiment"}[5m]))`

With this workaround you would achieve this instead with the following metrics query config:

```
        "query": {
          "type": "prometheus",
          "metricName": "statuses",
          "scopeLabel": "kubernetes_name=",
          "functions": ["irate::[5m]", "sum::"],
          "customInlineTemplate": "code=\"404\"",
          "serviceType": "prometheus"
        },
```

We are using `scopeLabel` and `locationLabel` (not in example) to define the filters set from `scope` and `location` respectively.

We also added `functions` to allow building a PromQL query using functions.  It is parsed as `<functionName>:<beginning args>:<ending args>`.

I'm not necessarily advocating  having this merged.  But I'm trying to bring visibility to issues when trying to use Prom metrics in kayenta's current state.

IMO the real fix is having orca `c.n.s.o.p.u.ContextParameterProcessor` properly evaluate the expression.  It seems to me it doesn't even need to parse it, just pass it through to be consumed/processed by the prometheus service directly.

### DIFF
--- a/kayenta-prometheus/src/main/java/com/netflix/kayenta/canary/providers/metrics/PrometheusCanaryMetricSetQueryConfig.java
+++ b/kayenta-prometheus/src/main/java/com/netflix/kayenta/canary/providers/metrics/PrometheusCanaryMetricSetQueryConfig.java
@@ -45,6 +45,15 @@ public class PrometheusCanaryMetricSetQueryConfig implements CanaryMetricSetQuer
   @Getter
   private List<String> groupByFields;
 
+  @Getter
+  private List<String> functions;
+
+  @Getter
+  private String scopeLabel;
+
+  @Getter
+  private String locationLabel;
+
   /**
    * @deprecated Use customInlineTemplate instead.
    */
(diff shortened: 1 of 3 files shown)


maggieneterval:
@duftler curious for your thoughts on this! Looks like Orca is possibly trying to evaluate ${}-wrapped expressions meant for Prometheus as SpEL?

fieldju:
As per @rarneson and @jmccann on the slack channel, this pr was to bring attention to the issue and wasn’t meant for merging.

https://github.com/spinnaker/kayenta/pull/579 is the PR that seems to fix the real issue here.

Closing this PR, re open if I am mistaken.

PR #576, opened 2019-06-27
feat(aws): support explicit AWS credentials

Aloren:
Add one more option for specifying AWS credentials.
https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-explicit


fieldju:
What is the use case that this solves that the default AWS credentials provider and the system property provider and environmental variable property provider doesn’t support?

Aloren:
@fieldju we would like to use minio as a configuration storage, it is amazon s3 compatible, but we do not use amazon infrastructure. This code gives an ability to configure credentials without any amazon specific configuration.

PR #577, opened 2019-06-27
feat(kayenta): Provide autoconfigurations for all Kayenta integration modules

Aloren:


Aloren:
We would like to setup Kayenta locally with custom configurations and modules added. We need to add only the required dependencies for our setup, like: `kayenta-web`, `kayenta-prometheus`, `kayenta-standalone`, but current project structure requires us to have all the dependencies making final jar too fat. Current approach gives possibility to add needed integration modules when required only (for custom build).

PR #582, opened 2019-07-01
fix(kayenta): Fixes generated pom.xml to omit invalid declaration of lombok

Aloren:
Fixes generated pom.xml to omit invalid declaration of lombok dependency.

compileOnly dependency should not be added to generated pom.xml,
see discussion: https://discuss.gradle.org/t/publishing-plugin-should-respect-compileonly-configuration/22903/2

Current `pom.xml` of any kayenta-* dependency is broken and leads to the following output when is used in Maven project:
```
[WARNING] The POM for com.netflix.kayenta:kayenta-web:jar:2.2.1 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
```
`kayenta-web` pom.xml contains the following invalid declaration:
Screen Shot 2019-07-01 at 8 43 06 PM
The issue with publishing compileOnly dependencies was fixed in latest `nebula-publishing-plugin` (https://github.com/nebula-plugins/nebula-publishing-plugin/commit/a5432aadc4b2886f1d9965361054d9614737c7df).
This PR contains override for the plugin version (it should be omitted as soon as `gradle-netflixoss-project-plugin` gets updated with latest nebula plugins).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,9 @@ buildscript {
     classpath "com.netflix.spinnaker.gradle:spinnaker-dev-plugin:$spinnakerGradleVersion"
     if (Boolean.valueOf(enablePublishing)) {
       classpath "com.netflix.spinnaker.gradle:spinnaker-gradle-project:$spinnakerGradleVersion"
+// TODO: nebula-publishing-plugin version override should be removed as soon as spinnaker-gradle-project is updated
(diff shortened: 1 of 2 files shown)


Aloren:
Also related to:
https://github.com/spinnaker/kork/pull/329
https://github.com/spinnaker/orca/pull/3017
as the same gradle configuration is used there.

PR #592, opened 2019-07-18
feat(deps): Depend on aws-java-sdk-s3 dependency instead of aws-java-sdk

Aloren:
I've looked through Kayenta dependencies and found that `aws-java-sdk` dependency contains 180 dependencies transitively. (see: https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk/1.11.592) According to my knowledge the only dependency that is required is `aws-java-sdk-s3` when using S3 storage. Currently `aws-java-sdk` is required, because `awsobjectmapper` creates `ObjectMapperConfigurer` that configures mapping of all aws models to/from json. I haven't found where it is actually used, so I thought that is not needed in Kayenta.


kayenta-s3/src/main/java/com/netflix/kayenta/s3/config/S3Configuration.java
```diff
@@ -43,8 +41,6 @@
   @DependsOn({"registerAwsCredentials"})
   public S3StorageService s3StorageService(
       AccountCredentialsRepository accountCredentialsRepository) {
-    AmazonObjectMapperConfigurer.configure(kayentaObjectMapper);
```
fieldju: Are we sure this isn't setting any Jackson options that the existing data relies on. 
Is there way to test that removing this line doesn't prevent previously written data to S3 from being deserialized?

fieldju: Here is what that method is currently setting.
``` java
    objectMapper.configure(MapperFeature.AUTO_DETECT_IS_GETTERS, false);
    objectMapper.configure(SerializationFeature.INDENT_OUTPUT, true);
    objectMapper.configure(SerializationFeature.WRITE_NULL_MAP_VALUES, false);
    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
Nothing there is jumping out like it could cause deserialization errors.

If you wanted to be paranoid you could add 
``` java
    objectMapper.configure(SerializationFeature.INDENT_OUTPUT, true);
    objectMapper.configure(SerializationFeature.WRITE_NULL_MAP_VALUES, false);
    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
    objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
To make sure everything behaves the same.

fieldju: https://github.com/spinnaker/kayenta/blob/master/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaConfiguration.java#L111

I think since that is explicitly set already that this change should be safe.

fieldju: I manually tested via Minio that the various objects that where written before this commit could still be deserialized.

PR #593, opened 2019-07-18
feat(standalone-canary-analysis): add SCAPE complete event listener and archive the status response object in the object store.

fieldju:
This is basically https://github.com/spinnaker/kayenta/pull/486 for the `/standalone_canary_analysis` endpoint so that the data needed for reporting is backed up outside of Redis and in the object store so that historical reporting will continue to work in the case of a Redis disaster.

I had to move the API domain models into core because the following is an enum rather than an interface: [ObjectType](https://github.com/spinnaker/kayenta/blob/master/kayenta-core/src/main/java/com/netflix/kayenta/storage/ObjectType.java#L29)

If it is preferred I can make a separate PR to convert that enum and its uses into an interface / impl's. rather than forcing all objects that can be stored be part of the core package.

in the process of moving the models I renamed `canaryanalysis` -> `standalonecanaryanalysis` to match the module name and hopefully avoid confusion.

To test I ran minio locally and used postman to trigger a SCAPE and verifyied to contents of the object store locally.

image

Then I killed the Redis DB and asserted that a GET to `standalone_canary_analysis/${ID}` still returned the payload as expected


kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/service/CanaryAnalysisService.java
```diff
@@ -100,10 +113,32 @@ public CanaryAnalysisExecutionResponse initiateCanaryAnalysisExecution(
...
+              storageAccountName,
+              AccountCredentials.Type.OBJECT_STORE,
+              accountCredentialsRepository);
+
+      StorageService storageService =
+          storageServiceRepository
+              .getOne(resolvedStorageAccountName)
+              .orElseThrow(
+                  () ->
+                      new IllegalArgumentException(
```
marchello2000: this changes the behavior from before, where if the execution wasn't found `ExecutionNotFoundException` would bubble up but now it will throw `IllegalArg` - not sure if other down stream consumers depend on the type of exception thrown (or am I missing something?)
Otherwise, i guess I would expect to only try to lookup execution if a storage account is configured (again, possible missing something since I am very new to this codebase)

marchello2000: similar to https://github.com/spinnaker/kayenta/pull/593/files#diff-729f4e7ba6697e7b7d60f05de492e912R35

fieldju: >I would expect to only try to lookup execution if a storage account is configured

I think this is reasonable, I will adjust the logic to clean this up.

kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/service/CanaryAnalysisService.java
```diff
@@ -100,10 +113,40 @@ public CanaryAnalysisExecutionResponse initiateCanaryAnalysisExecution(
...
+                                        "No storage service was configured for %s unable to load archived results.",
+                                        resolvedStorageAccountName)));
+
+                return (CanaryAnalysisExecutionStatusResponse)
+                    storageService.loadObject(
+                        resolvedStorageAccountName,
+                        ObjectType.STANDALONE_CANARY_RESULT_ARCHIVE,
+                        canaryAnalysisExecutionId);
+              })
+          .orElseThrow(() -> e);
```
fieldju: @marchello2000 What do you think about that?

It throws the original error if you didn't add ?storageAccount=xxxx to the request.

It throws an Illegal Argument Exception or other downstream exceptions from `storageService.loadObject` if you supply that new optional query param and it doesn't load.

marchello2000: 👍

PR #594, opened 2019-07-19
feat(kayenta): Prepare automated integration tests for kayenta

Aloren:
Simple successfull and failure cases added based on prometheus and s3 integrations


Aloren:
@fieldju @cfieber @skandragon Please review. These tests are completely automated and require only docker to be present on machine to be run. 
Dependencies `embedded-redis` and `embedded-minio` were initially developed by Playtika, but have now migrated to `testcontainers` repository (https://github.com/testcontainers/testcontainers-spring-boot), I hope there are no issues with external deps.

kayenta-integration-tests/src/test/java/com/playtika/services/infra/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.core.env.MapPropertySource;
+import org.testcontainers.Testcontainers;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
+import org.testcontainers.utility.MountableFile;
+
+@Slf4j
+@Configuration
+public class EmbeddedPrometheusBootstrapConfiguration {
```
vasilievip: please add additional submodule to testcontainers-spring-boot

Aloren: okay

Aloren:
> Change the package names to match the rest of the project (com.netflix.kayenta....)

Sorry, I was copypasting from our internal project :laughing: Fixed already.

fieldju:
I have a couple comments / thoughts.

1. Does it make more sense to have integration tests live with the modules they tests as [a new source set](https://github.com/spinnaker/kayenta/blob/bf167627d51c0970ade10ce60583a2587cdb012f/kayenta-signalfx/kayenta-signalfx.gradle#L16) such as done already in this project with [Graphite](https://github.com/spinnaker/kayenta/tree/bf167627d51c0970ade10ce60583a2587cdb012f/kayenta-graphite/src/integration-test) and [SignalFX](https://github.com/spinnaker/kayenta/tree/bf167627d51c0970ade10ce60583a2587cdb012f/kayenta-signalfx/src/integration-test)

My personal opinion is that that the Prometheus test code belongs in the Prometheus module and the S3 test code belongs in the S3 module, etc.

2. We should align on a method for embedding Redis.

I think I have seen something about embedded Redis in some Kork stuff, Graphite and SignalFx integrations are using this [embedded-redis](https://github.com/kstyrc/embedded-redis) plugin, and this PR would introduce a 3rd way of doing it.

fieldju:
> I suppose that travis setup should also be updated, but I haven't yet figured out where it can be configured in the repo.

@Aloren 

You configure this stuff in the [travis.yaml file](https://github.com/spinnaker/kayenta/blob/master/.travis.yml)

You would add a block along the lines of this

```yaml
after_success:
# run the integration test for Pull requests
- test "${TRAVIS_PULL_REQUEST}" != "false" && ./gradlew integrationTest --info
```

Aloren:
@fieldju I've already thought regarding this. There are a couple of points why I've done it this way:
1. If we have integration tests split over all modules -- we will probably have to setup kayenta multiple times during build -- each module will require startup of kayenta and embedded resources. 
In case of one integration-tests module -- we can setup Kayenta and all required resources only once. 
2. With one integration-tests module we can also check different integrations together.
3.  In current example I've setup only Prometheus and Minio, but it doesn't mean that we can't setup signalfx or influx or other integrations.
4. `kayenta-web` is an application that provides multiple integrations out of the box that must be checked. **Currently there are no automated tests that can be run after introducing new changes.** You don't really know whether kayenta-s3 will startup when changes into kayenta-prometheus are introduced.
5. Some generic test cases that test Kayenta internal flows or management endpoints should be also checked. Should we check it in kayenta-prometheus or kayenta-influx? Or should we check it in both?
6. > I think I have seen something about embedded Redis in some Kork stuff, Graphite and SignalFx integrations are using this embedded-redis plugin, and this PR would introduce a 3rd way of doing it.

Current approach setups an environment required for the tests automatically and requires only docker to be present on the machine -- I personally think this is much more developer friendly than having Redis present on the machine. :)

fieldju:
> @fieldju I've already thought regarding this. There are a couple of points why I've done it this way:
> 
> 1. If we have integration tests split over all modules -- we will probably have to setup kayenta multiple times during build -- each module will require startup of kayenta and embedded resources.
>    In case of one integration-tests module -- we can setup Kayenta and all required resources only once.
> 2. With one integration-tests module we can also check different integrations together.
> 3. In current example I've setup only Prometheus and Minio, but it doesn't mean that we can't setup signalfx or influx or other integrations.
> 4. `kayenta-web` is an application that provides multiple integrations out of the box that must be checked. **Currently there are no automated tests that can be run after introducing new changes.** You don't really know whether kayenta-s3 will startup when changes into kayenta-prometheus are introduced.
> 5. Some generic test cases that test Kayenta internal flows or management endpoints should be also checked. Should we check it in kayenta-prometheus or kayenta-influx? Or should we check it in both?
> 6. > I think I have seen something about embedded Redis in some Kork stuff, Graphite and SignalFx integrations are using this embedded-redis plugin, and this PR would introduce a 3rd way of doing it.
> 
> Current approach setups an environment required for the tests automatically and requires only docker to be present on the machine -- I personally think this is much more developer friendly than having Redis present on the machine. :)

Shouldn't you be porting the SignalFx and Graphite integration tests to this new module if you want to change the way Integration tests are done as part of this PR?

Aloren:
> Shouldn't you be porting the SignalFx and Graphite integration tests to this new module if you want to change the way Integration tests are done as part of this PR?
Sorry, but I don't see why this is connected. You've added the tests for the integrations you are using in your company, I've added the tests for the integrations we are using in our company. Prometheus and S3 modules shouldn't be tested? :)

fieldju:
> > Shouldn't you be porting the SignalFx and Graphite integration tests to this new module if you want to change the way Integration tests are done as part of this PR?
> 
> Sorry, but I don't see why this is connected. You've added the tests for the integrations you are using in your company, I've added the tests for the integrations we are using in our company. Prometheus and S3 modules shouldn't be tested? :)

Correct me if I am misunderstanding you. I think you have asserted the following.

1. All integration tests should live in a new integration test module that you are introducing that is separate from the code that they test.
1. The mechanism for embedding Redis for said integration tests should be your newly introduced containerized solution.
1. We should enable your new integration test module to be ran automatically for new changes via Travis.

Assuming the above then I think you should be moving / uplifting the existing integration tests as part of this PR.

Aloren:
If you agree on the current approach — moving existing tests is not a problem for me.

fieldju:
> If you agree on the current approach — moving existing tests is not a problem for me.

I am fine with either as long as there is consistency (which is why I am advocating the uplifting of the existing tests).

We have to wait and see what @duftler and @cfieber thoughts are, since they would be the gate keepers on this right now.

fieldju:
Logistically, I have a couple thoughts on potential gotcha's.

- The SFX integration tests require an API key. I am reaching out to SignalFx to see if I can get one for this project that we don't pay for.
- The SignalFx and Graphite integration tests generate metrics for the tests via these [TestConfiguration Spring Components](https://github.com/spinnaker/kayenta/blob/master/kayenta-graphite/src/integration-test/java/com/netflix/kayenta/config/GraphiteIntegrationTestConfig.java#L36)
  - Basically they create a thread for each mock instance, you need at least 3 threads per metric source.
    - **Will this be an issue in travis if we have 3-6 threds running per metric source integration**?

kayenta-integration-tests/src/test/resources/prometheus.yml
```diff
@@ -0,0 +1,8 @@
+global:
+  scrape_interval: 1s
+
+scrape_configs:
+  - job_name: kayenta-test-metrics
+    metrics_path: /prometheus
+    static_configs:
+      - targets: ['host.testcontainers.internal:8081']
```
bsideup: Wow, nice :) When I was adding this feature, I was thinking about Selenium & friends, but I'm happy to see it used for other cases too  👍

Aloren: @bsideup I was really glad to find this feature! Thank you! `host.docker.internal` as you know is available only on mac and windows :(

bsideup: Even if they add it to the native Docker, sometimes your "localhost" is not Docker's host (think remote Docker, or when you're running your tests inside a container already)

Testcontainers' approach works everywhere because we start a reverse port forwarding SSH session to the SSH server running as a container inside the target Docker environment :)

fieldju:
I need to make some modifications to the New Relic integration and want to write some integration tests for that metric source.

I am going to attempt do that using this branch / the patterns laid out here and sanity check that they are generic / extensible.

fieldju:
OK, I just checked out this branch and took a deeper look into it.

@Aloren, **Is this PR supposed to represent a pattern for generically testing Kayenta or is this supposed to be a single set of integration tests that are specific to just your use case (prometheus + S3 via Minio)?**

I was under the impression that you are trying to establish a new pattern via this single module that is for testing all current and future metric source / storage integrations.

If that is the case, I am seeing a few issues that prevent this from being generic / extesible for use by other metric source integrations.

1. In the [CanaryConfigReader class](https://github.com/spinnaker/kayenta/blob/418e1e7ae56272d1ea52bb46be0b4de3e2154872/kayenta-integration-tests/src/test/java/com/netflix/kayenta/utils/CanaryConfigReader.java#L31) you are configuring an Object mapper that will only work with Prometheus rather than getting the [Kayenta Object mapper that has been configured to support all metric source integrations](https://github.com/spinnaker/kayenta/blob/master/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaConfiguration.java#L98).
2. The [StandaloneCanaryAnalysisSteps](https://github.com/spinnaker/kayenta/blob/418e1e7ae56272d1ea52bb46be0b4de3e2154872/kayenta-integration-tests/src/test/java/com/netflix/kayenta/steps/StandaloneCanaryAnalysisSteps.java#L51-L52) is hard coded to Prometheus / Minio and should likely ultimatly source those variables from the [application-casses.yml](https://github.com/spinnaker/kayenta/blob/418e1e7ae56272d1ea52bb46be0b4de3e2154872/kayenta-integration-tests/src/test/resources/application-cases.yml)
3. The classname and or test names in [StandaloneCanaryAnalysisTest.java](https://github.com/spinnaker/kayenta/blob/418e1e7ae56272d1ea52bb46be0b4de3e2154872/kayenta-integration-tests/src/test/java/com/netflix/kayenta/StandaloneCanaryAnalysisTest.java#L26) are nammed generically but they really just a set of tests that verify that Prometheus and Minio work with that endpoint.
    - How is this meant to be extended to support additional metric sources?
        - Is a contributor expected make a new class that extends the `BaseIntegrationTest` class? In that case this class name should be renamed.
        - Is a contributor expected to add additional methods to this class? In that case the test names should be renamed.
4. It is unclear how [MetricsReportingConfiguration](https://github.com/spinnaker/kayenta/blob/418e1e7ae56272d1ea52bb46be0b4de3e2154872/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/MetricsReportingConfiguration.java#L32) class is configured to generically report metrics to N metrics providers or how a contributor goes about setting that up.
    - I assume that it is using some Spring magic with the following dependencies
        - testCompile 'io.micrometer:micrometer-registry-prometheus
<br/>The [micrometer.io website](https://micrometer.io/) states that it supports all the current integrations with Kayenta which is good. ~~Does it support sending metrics to multiple sources? If it doesn't then it defeats the purpose of bootstrapping the Kayenta service once per set of tests.~~ 
  <br/>Micrometer supports [composite registries](https://micrometer.io/docs/concepts#_composite_registries) which allow it to send metrics to more than one monitoring system.
<br/>I think a composite registry should be getting bootstrapped in that class with the prometheus registry added to it.

Aloren:
For making this approach generic there needs to be done some changes. If everyone agrees using this approach I will rework PR.

Aloren:
Just let me know whether gatekeepers are ok with that and whether s3 and graphite should be included into this PR. Thanks

fieldju:
re @Aloren 
> For making this approach generic there needs to be done some changes. If everyone agrees using this approach I will rework PR.

I talked to @cfieber on slack, we I think we concluded the single module approach for housing all the integration tests is a good pattern.

I think there are still questions on how the process should work for each metric source / storage integration, etc.

For example you had stated

>In case of one integration-tests module -- we can setup Kayenta and all required resources only once.

@cfieber expressed concerns that it might not be possible to do this, but I am pretty sure that you can for the modules we are talking about as each metric source / storage account basically just becomes a named account via `/credentials` that has a type of `METRICS_STORE, CONFIGURATION_STORE, or OBJECT_STORE` and you just select the SUT via the metricsAccount, configurationAccount and storageAccount query params.

@cfieber then stated: 

> in that case then it seems reasonable to not worry too much about what is enabled or disabled - I still don’t totally know about just starting up kayenta once and hoping that none of the tests leave the container in a weird state or have side effects that might affect the other tests

There are concerns about unintended side effects of not reseting the state of the application between each test suite.

I am not 100% sure either way.

What is the process / pattern are thinking @aloren?

```
1. Start Kayenta and Redis using a single mega configured kayenta.yml
2. Run various other tests
2. Run /canary for each configured metric source (maybe in parrellel)
3. Stop Kayenta
```

vs

```
0. Repeat steps 1-3 for each Integration
1. Start Kayenta and Redis using a specific kayenta.yaml for the given test suite (Prometheus, SignalFx, etc)
2. Run the tests
3. Shutdown Kayenta
```

or something else?

> Just let me know whether gatekeepers are ok with that and whether s3 and graphite should be included into this PR. Thanks

I think once we agree on the general pattern (started Kayenta once, vs starting it for each integration), if you ported Graphite to prove that the pattern works for n+1 (Prometheus and Graphite), I can then make a new PR after this gets merged uplifting the SignalFx integration tests to the new pattern.

Aloren:
> I am pretty sure that you can for the modules we are talking about as each metric source / storage account basically just becomes a named account

Yes, that's what I was thinking about. That it is possible to setup all integrations and use different accounts for different tests.

> There are concerns about unintended side effects of not reseting the state of the application between each test suite.

Hm.. I think I don't understand what side effects might be? Regarding metrics storage --  different tests need to use different metrics + set of tags (if applicable), so that these tests do not depend on each other. Regarding object/configuration storage -- data is stored per `pipelineExecutionId`, so there shouldn't be any issues with any side effects.

> What is the process / pattern that you are thinking @Aloren?

I personally prefer the first pattern. Start -> run basic checks -> run different canary analyses -> stop.
It can still be changed in future if this will be required. (Maybe just mark tests with `@DirtiesContext`).

> if you ported Graphite to prove that the pattern works for n+1 

I will add Graphite integration into this PR as soon as I get some free time, sorry :) (maybe on friday or weekdays)

Aloren:
@fieldju @cfieber I've added tests for graphite and in-memory store integration. Please have a look.

Aloren:
Any chances these changes can be reviewed?

fieldju:
Every looks fine to me code wise.

My last comment which I hope someone can chime in on is what the preference is for Spinnaker projects is in terms of integration tests being opt-out vs opt-in.

In projects that I control, I never have integration tests, that are some times fragile and / or long running, run with the unit tests.

I normally create a new source set for my JVM projects and explicitly opt-in to them via setting up CI or running them on my machine via `./gradlew integration` or `./gradlew integration-test`.

Right now if I run the `check` task on my machine it takes a long time to run, I can do the following `./gradlew clean check -x xxxxxx:check` to opt-out, but I wonder if they should be opt-in instead.

kayenta-integration-tests/kayenta-integration-tests.gradle
```diff
@@ -0,0 +1,15 @@
...
+    testCompile project(":kayenta-web")
+    testCompile 'io.rest-assured:rest-assured:3.1.1'
+    testCompile 'org.awaitility:awaitility:3.1.6'
+    testCompile 'io.micrometer:micrometer-registry-prometheus'
+    testCompile 'io.micrometer:micrometer-registry-graphite'
+    testCompile 'org.springframework.cloud:spring-cloud-starter:2.1.2.RELEASE'// needed for bootstrap phase when all embedded containers are setup
+    testCompile 'com.playtika.testcontainers:embedded-redis:1.25'
+    testCompile 'com.playtika.testcontainers:embedded-minio:1.25'
+}
+
```
fieldju:
```groovy
gradle.taskGraph.whenReady {
    tasks.test.enabled = System.getProperty('ENABLE_INTEGRATION_TESTS', 'false') == 'true'
}
```

Can we throw this somewhere in this file, to make the integration tests opt in.

PR #598, opened 2019-07-24
fix(typeahead/prometheus): Consider account name when querying metadata.

duftler:


maggieneterval:
@spinnakerbot cherry-pick 1.15

PR #599, opened 2019-07-24
chore(intellij): Add IDEA project files.

plumpy:


fieldju:
I've never seen projects include the IDE specific files, they are usually on the git ignore list.

What is the benefit of including these?
Will these trample my existing settings when I pull next?

The copyright profile you are including is hard coded to Netflix, I was under the impression that as contributors add new files that the copyright notice should have their org in the header.

plumpy:
It makes things a lot easier for new developers who can just open the project and have a working IntelliJ setup, versus having to go in and configure things the way we require.

It will overwrite the following settings first time you pull past this commit (specifically, git will overwrite your local files with these files). After that, you can make any changes you want to them and they won't be overwritten again.

The included settings are:
* Use Git for VCS
* Use the gradle wrapper (instead of a local gradle install)
* Turn on gradle auto-import
* Enable the google-java-format plugin (if installed)
* Turn on annotation processing
* Add copyright headers to new files

You can change the company in the copyright header by adding an `io.spinnaker.copyright` property to your `~/.gradle/gradle.properties`.

We've had these files checked in to the `clouddriver` and `echo` repositories for several weeks now and no one has complained about any issues yet. (I intentionally let them sit around for a while before moving to commit them to other projects...)

fieldju:
> You can change the company in the copyright header by adding an io.spinnaker.copyright property to your ~/.gradle/gradle.properties.

I will do that.

> Enable the google-java-format plugin (if installed)

I will look into this, does this behave differently than the spotlessApply gradle task that is configured to run as a git hook in this project?

plumpy:
> I will look into this, does this behave differently than the spotlessApply gradle task that is configured to run as a git hook in this project?

It behaves the same, but means you can use Code→Reformat Code to format the code properly, rather than waiting on a gradle to do it...

Issue #595, opened 2019-07-22
[RFC] Proposed updates to the New Relic Insights metrics configuration to make the user experience less confusing.


fieldju:
The current way of configuring a New Relic Insights can be bit confusing for users to figure out.

Assume I have the following scope params

- scope: `myservice-prod-v01` 
- location: `us-west-2`

and the following extended Scope Parameters:

| Scope Key  | Scope Value |
| ------------- | ------------- |
| `_scope_key`  | asg_name |
| `_location_key`  | region |

Assume I want to create a New Relic insights query that gets the number of server errors for the following uri `/foo`, I would configure the following query for a metric as follows.

 
![image](https://user-images.githubusercontent.com/711726/61671371-31175880-ac9c-11e9-8e5e-d6a26b87ccbd.png)

```json
{
    "query": {
        "type": "newrelic",
        "select": "SELECT count(*) FROM Transactions",
        "q": "httpStatusCode like '5%' AND uri like '/foo'"
    }
}
```

This will produce the following query

```
SELECT count(*) 
FROM Transactions 
WHERE httpStatusCode like '5%' 
  AND uri like '/foo' 
  AND asg_name LIKE 'myservice-prod-v01' 
  AND region like 'us-west-2'
```

I think the original intent / goal was to have a flexible way of defining the queries that abstracted away NRQL.
It is a bit confusing to figure out how the pieces stitch together, because you still have to write out partial NRQL.

I propose we do the following (I am committing to do the work and make a PR).

I would like to essentially make this integration work more like the Prometheus integration where you can have inline templates or a simple abstracted user experience.

I essentially want the user experience in the UI to be choose from the following radio options

```
Select your query type:
- ( ) Simple
- ( ) Inline Template
```

If you select `Inline Template` the UI updates to just a single box labeled `NRQL` and where you could type something like the following 

```
SELECT count(*) 
FROM Transactions 
WHERE httpStatusCode like '5%' 
  AND uri like '/foo' 
  AND asg_name LIKE '${scope}' 
  AND region like '${location}'
```

![image](https://user-images.githubusercontent.com/711726/61671339-fdd4c980-ac9b-11e9-85fa-a2fcdcd12507.png)

This generates the following JSON

```json
{
    "query": {
      "type": "newrelic",
      "queryType": "inlineTemplate",
[...]

PR #603, opened 2019-07-29
feat(newrelic): default scope keys and inline template support

fieldju:
Closes: https://github.com/spinnaker/kayenta/issues/595

This PR adds the ability for operators of Kayenta to set a default scope key and scope location key when configuring a New Relic account.

This PR also adds inline template support to the New Relic integration.

See: https://github.com/Nike-Inc/kayenta/blob/feature/newrelic_inline_templates/kayenta-newrelic-insights/src/test/groovy/com/netflix/kayenta/newrelic/metrics/NewRelicQueryBuilderServiceSpec.groovy#L99-L131 for inline template example via the unit tests.

The only testing I have done is the unit tests for the query building logic. I still need to do some manual UAT'ing to make sure everything still functions.

This PR should be 100% backwards compatible because all the changes where additive.

Also the `select plus q` config flow can be used to enable the simple flow in the UI that is described in https://github.com/spinnaker/kayenta/issues/595

CC: 
@imakewebthings
@ehmke


kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/canary/providers/metrics/NewRelicCanaryMetricSetQueryConfig.java
```diff
@@ -24,22 +24,40 @@
...
+
   @Getter private String q;
 
   @NotNull @Getter private String select;
 
   @Getter private String resolution;
 
+  @Getter private String customInlineTemplate;
+
+  @Getter private String customFilterTemplate;
```
fieldju: I added this because this is how the system that @duftler made for templates works, but I am not 100 sure how filters work and if they are relevant to New Relic.

fieldju:
One issue that might crop up is currently users can write invalid NRQL and valid NRQL that returns invalid results.

fieldju:
OK, I UAT'd things and found an issue that I have now fixed (I needed to expose the start and end time as template keys for users).

I updated the NR fetch controller to allow me to do some UAT'ing and ran the following

#### I proved that using select and q worked as they did before to maintain backwards compatibility

<details><summary>Screen Shots</summary><p>



</p> 
</details>

#### I Proved that I could set a default scope key and location key and if I omitted them from the scope parameters that the query generator would use the defaults.

fieldju:
Added one last tweak, which is to add the step parameter to the filter keys.

ex: 

```json
"customInlineTemplate": "SELECT count(*) FROM Transaction TIMESERIES ${step} seconds SINCE ${startEpochSeconds} UNTIL ${endEpochSeconds} WHERE status_code LIKE '5%' AND asg_name LIKE '${scope}' AND region LIKE '${location}' AND foo LIKE ${foo}"
```
which generates the following

```sql
SELECT count(*) FROM Transaction TIMESERIES 60 seconds SINCE 1564358400 UNTIL 1564444800 WHERE status_code LIKE '5%' AND asg_name LIKE 'myService' AND region LIKE 'us-west-2' AND foo LIKE bar
```

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/canary/providers/metrics/NewRelicCanaryMetricSetQueryConfig.java
fieldju: @duftler is this specific to prometheus and or stack driver, or would this be relevant here too?

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/newrelic/metrics/NewRelicQueryBuilderService.java
```diff
@@ -0,0 +1,183 @@
...
+    // You cannot use the keys start and end here as the template engine tries to read the value out
+    // of the canary
+    // scope bean instead of the values you supply here and you get a casting error.
+    // Thus we add the epoch millis to the context to be available in templates as startEpochSeconds
+    // and endEpochSeconds
+    Map<String, String> originalParams =
+        canaryScope.getExtendedScopeParams() == null
+            ? new HashMap<>()
+            : canaryScope.getExtendedScopeParams();
+    Map<String, String> paramsWithExtraTempalteValues = new HashMap<>(originalParams);
```
ehmke: typo

kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
```diff
@@ -147,7 +147,11 @@ private static void populateTemplateBindings(
           }
         }
 
-        String propertyValue = (String) propertyDescriptor.get().getReadMethod().invoke(bean);
+        // Some stuff like start, end and step can't be cast to a string.
+        String propertyValue =
```
fieldju: @duftler this is the only change that could effect other integrations, I think it is safe though.

duftler: This change seems safe to me as well.

kayenta-newrelic-insights/src/main/java/com/netflix/kayenta/canary/providers/metrics/NewRelicCanaryMetricSetQueryConfig.java
duftler: `customFilterTemplate` refers by name to a template contained in the templates map on the canary config. `customInlinetemplate` directly contains the actual template to use. The prometheus canary config metric modal allows one to configure inline templates.

I think it's relevant for this provider as well.

PR #604, opened 2019-08-02
feat(canary): Add additional metadata to the canary status response.

fieldju:
This PR adds the metric account used for the canary to the canary status response, so that we can have a modify and retry button in our retrospective analysis tool / reports.

I also did some misc clean up in the Execution Mapper for the canary status response. Mainly extracting some duplicated code into a method and converting some if then blocks into optional chains.

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/ExecutionMapper.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/ExecutionMapper.java
@@ -15,6 +15,8 @@
  */
 package com.netflix.kayenta.canary;
 
+import static java.util.Optional.ofNullable;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


kayenta-core/src/main/java/com/netflix/kayenta/canary/ExecutionMapper.java
```diff
@@ -89,104 +81,66 @@ public ExecutionMapper(
...
-      canaryExecutionStatusResponseBuilder.canaryConfigId(canaryConfigId);
-    }
-    if (contextContext.containsKey("configurationAccountName")) {
-      String configurationAccountName = (String) contextContext.get("configurationAccountName");
-      canaryExecutionStatusResponseBuilder.configurationAccountName(configurationAccountName);
-    }
+                .pipelineId(pipeline.getId());
+
+    pipeline.getStages().stream()
+        .filter(stage -> stage.getName().startsWith("fetch"))
```
marchello2000: not super familiar with the inner workings/usage of orca here but looking for stage with a given name seems brittle... can you key off stage type here?

marchello2000:
the refactor looks great, thank you!
a bit concerned about the stage lookup by name (https://github.com/spinnaker/kayenta/pull/604/files#diff-d16c385276e2c78b50810b445c82e426R116) but i don't know the details of how orca is used under the covers.. just seems brittle.
Otherwise LGTM

kayenta-core/src/main/java/com/netflix/kayenta/canary/ExecutionMapper.java
fieldju: Yes, I think you are correct. 

Even better I think I will update the PR to just store this piece of information in the setupContext stage rather than even needed to rely on filtering one of the fetch stages.

fieldju:
> the refactor looks great, thank you!
> a bit concerned about the stage lookup by name (https://github.com/spinnaker/kayenta/pull/604/files#diff-d16c385276e2c78b50810b445c82e426R116) but i don't know the details of how orca is used under the covers.. just seems brittle.
> Otherwise LGTM

I updated the PR to just store the metrics account name with the rest of the context for the over all pipeline.

Speaking of brittle things, I am not a huge fan of the generic maps and would love to create models for everything with JS308 validation and use Jackson to convert back and forwarth 

ex:

```java
Map<String, Object> map = 
    mapper.convertValue(foo, new TypeReference<Map<String, Object>>() {});
```

```java
final MyPojo pojo = mapper.convertValue(map, MyPojo.class);
```

PR #609, opened 2019-08-30
refactor(newrelic): get nr endpoint from config

derekmiller:
There's a couple of reasons for this: 1) we need to be able to configure this endpoint for staging, 2) this is the convention that the other providers follow; there's no reason to have this be hardcoded as a constant and not configurable.


derekmiller:
Okay @dotdotdotpaul - updated to set a default if not set in the config.

Issue #610, opened 2019-09-05
FR: Support submitting a readable canary config Id on initial save.


jtk54:
The canary config UUIDs aren't very clear when codifying pipelines and canary configs. I would love it if either 1. you could submit a readable canary config id in the payload and use that to reference the config or 2. you could reference the canary config by name.

PR #611, opened 2019-09-06
feat(config): Allow canary config id to be passed in instead of alway…

duftler:
…s randomly generated.

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryConfig.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryConfig.java
@@ -39,6 +39,8 @@ public class CanaryConfig {
 
   @NotNull @Getter @Setter private String name;
 
+  @Getter private String id;
+
   @NotNull @Getter private String description;
 
   @NotNull @Getter private String configVersion;
--- a/kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
+++ b/kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
@@ -111,14 +111,20 @@ public class CanaryConfigController {
     canaryConfig.setUpdatedTimestampIso(
         Instant.ofEpochMilli(canaryConfig.getUpdatedTimestamp()).toString());
 
-    String canaryConfigId = UUID.randomUUID() + "";
+    if (StringUtils.isEmpty(canaryConfig.getId())) {
+      // Ensure that the canary config id is stored within the canary config itself.
+      canaryConfig = canaryConfig.toBuilder().id(UUID.randomUUID() + "").build();
+    }
 
-    // TODO(duftler): Serialize the canary config within a canary run?
+    String canaryConfigId = canaryConfig.getId();
+
     validateNameAndApplicationAttributes(canaryConfig);
 
     try {
       configurationService.loadObject(
           resolvedConfigurationAccountName, ObjectType.CANARY_CONFIG, canaryConfigId);
+
+      throw new IllegalArgumentException("Canary config '" + canaryConfigId + "' already exists.");
     } catch (NotFoundException e) {
       configurationService.storeObject(
           resolvedConfigurationAccountName,
@@ -130,8 +136,6 @@ public class CanaryConfigController {
 
       return CanaryConfigUpdateResponse.builder().canaryConfigId(canaryConfigId).build();
     }
-
-    throw new IllegalArgumentException("Canary config '" + canaryConfigId + "' already exists.");
   }
 
   @ApiOperation(value = "Update a canary config")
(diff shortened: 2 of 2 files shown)


kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryConfig.java
```diff
@@ -39,6 +39,8 @@
 
   @NotNull @Getter @Setter private String name;
 
+  @NotNull @Getter @Setter private String id;
```
duftler: Had to add `id` to canary config, and it will be populated anytime the config is persisted.

It still does not _need_ to be provided on a POST or a PUT, so no changes are required to any dependent services.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
```diff
@@ -111,14 +111,26 @@ public CanaryConfigUpdateResponse storeCanaryConfig(
...
         Instant.ofEpochMilli(canaryConfig.getUpdatedTimestamp()).toString());
 
-    String canaryConfigId = UUID.randomUUID() + "";
+    String canaryConfigId;
+
+    if (StringUtils.isEmpty(canaryConfig.getId())) {
+      canaryConfigId = UUID.randomUUID() + "";
+
+      // Ensure that the canary config id is stored within the canary config itself.
+      if (StringUtils.isEmpty(canaryConfig.getId())) {
```
jtk54: This `if` seems unnecessary.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
```diff
@@ -111,14 +111,26 @@ public CanaryConfigUpdateResponse storeCanaryConfig(
...
+        canaryConfig = canaryConfig.toBuilder().id(canaryConfigId).build();
```
jtk54: Don't you have an `@Setter`?

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
duftler: Thanks, copy/paste error.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
duftler: Shouldn't, will remove it. Trying to stick to not modifying the instance once it's constructed.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
```diff
@@ -111,14 +111,24 @@ public CanaryConfigUpdateResponse storeCanaryConfig(
...
 
-    String canaryConfigId = UUID.randomUUID() + "";
+    String canaryConfigId;
+
+    if (StringUtils.isEmpty(canaryConfig.getId())) {
+      canaryConfigId = UUID.randomUUID() + "";
+
+      // Ensure that the canary config id is stored within the canary config itself.
+      canaryConfig = canaryConfig.toBuilder().id(canaryConfigId).build();
+    } else {
```
plumpy: Probably just personal preference, but I think this would read easier if you removed the else block, so ti's more like

```
if (isEmpty(canaryConfig.getId())) {
  canaryConfig = canaryConfig.toBuilder().id(UUID.randomUUID().toString()).build();
}

String canaryConfigId = canaryConfig.getId();
```

Feel free to ignore if you disagree.

:+1: to not modifying the object more than we already do... it's certainly surprising to pass an object to a `store()` method and find out the method secretly also modified it.

duftler: That does read easier. Changed it.

duftler:
Closes https://github.com/spinnaker/kayenta/issues/610.

csanden:
LGTM

kayenta-core/src/main/java/com/netflix/kayenta/canary/CanaryConfig.java
```diff
@@ -39,6 +39,8 @@
 
   @NotNull @Getter @Setter private String name;
 
+  @NotNull @Getter private String id;
```
fieldju: If this field is not null, then you should populate this field on the GET requests to configs that already exist but do not yet have the id saved as part of the object.

https://github.com/spinnaker/kayenta/blob/c2acf40a2752d8748ce466325f09445321ce6299/kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java#L61-L79

fieldju: Maybe this should be @nullable, so that way if JSR 308 validation is turned on people who don't supply an id, since it's technically optional won't get an API error.

duftler: Good point. Have removed `@NotNull`.

PR #617, opened 2019-09-27
feat(signalfx): Provide additional context about the results of the signalflow programs.

fieldju:
We noticed that in Signal Fx if you have metrics that are reporting at differing frequencies. 
Say a system agent on an EC2 instance that reports system cpu and memory at 1 data point per minute and application metrics that report 6 times per minute.
In Kayenta if you set the step to 10 (6 datapoints per minute) that the CPU and Memory data gets padded with a bunch of NaNs and gets graphed weird.

Just taking that data and stretching it had some side effects as well, so this PR adds more attributes to the data that flows through with the results so that the data can be reconstructed and graphed more accurately.

Before:

image

After:

image

CC: @melanahammel

### DIFF
--- a/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
+++ b/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Nike, inc.
+ * Copyright (c) 2019 Nike, inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
```diff
@@ -56,6 +56,16 @@
 @Slf4j
 public class SignalFxMetricsService implements MetricsService {
 
+  private static final String SIGNAL_FLOW_ERROR_TEMPLATE =
+      "An error occurred whole executing the Signal Flow program. "
```
nisanharamati: typo: `whole` -> `while`?

fieldju: https://github.com/spinnaker/kayenta/pull/618

PR #620, opened 2019-10-12
feat(secrets/gcs): Bumping orca dep for pulling gcs secrets module

german-muzquiz:
Bumping orca dependency to include new kork-secrets-gcp module.
PR introducing GcsSecretsEngine: spinnaker/kork#380

Fixes: spinnaker/spinnaker#4668


fieldju:
> The changes look reasonable to me.
> 
> Can you please indicate how you've tested kayenta to ensure it still functions properly with the latest orca version?
> 
> In the past, we have encountered subtle changes to orca/keiko that required kayenta changes when upgrading.

With @Aloren's change to the build process a set of integration tests now run on every PR.
So I can say that Kayenta still functions end to end with at least Graphite and Prometheus.

duftler:
> With @Aloren's change to the build process a set of integration tests now run on every PR.
> So I can say that Kayenta still functions end to end with at least Graphite and Prometheus.
> 
> https://travis-ci.org/spinnaker/kayenta/builds/597037661?utm_source=github_status&utm_medium=notification

Works for me.

LGTM.

PR #621, opened 2019-10-14
fix(judge): Disable first pass outlier removal by default.

duftler:
Adds the ability to re-enable outlier removal on a per-metric basis:
```
    {
      "name": "5xx rate",
      "query": {
        "type": "prometheus",
        "customInlineTemplate": "PromQL:avg(rate(gateway_http_responses{code=\"5xx\", pod=~\"$\\{scope}\"}[2m]))",
        "serviceType": "prometheus"
      },
      "groups": [
        "Group 1"
      ],
      "analysisConfigurations": {
        "canary": {
          "outlierStrategy":"remove"
        }
      },
      "scopeName": "default"
    },
```

PR #622, opened 2019-10-16
chore(docs): add documentation on the object structure of canary config.

fieldju:
As discussed here: https://github.com/spinnaker/kayenta/pull/621#discussion_r334932676

This is a PR of my internal documentation of the Object structure of the Canary Config that documents some of the hidden features that the UI doesn't expose.

These probably aren't perfect but are better than nothing.

One thing that I am still confused on that maybe @csanden can add input on is the difference between `critical` and `allowed` increase / decrease.


docs/canary-config.md
```diff
@@ -0,0 +1,99 @@
+Objects in this document are documented using [Markdown Syntax for Object Notation (MSON)].
+
+### Canary Config Object model (object)
+
+#### Properties
+- `name` **my-app golden signals canary config** (string, required) - Name for canary configuration.
```
duftler: Could also add the `id` since it can now be specified on the config when it is pushed.

nisanharamati: Not sure how deep this should go, but: in orca, at least one of id and name (but not both) is required. If both are provided, name is ignored and id is used.

fieldju: - [ ] Add id

docs/canary-config.md
```diff
@@ -0,0 +1,99 @@
...
+<a name="effect-size"></a>
+### EffectSize
+Controls the degree of statistical significance the metric needs to fail or fail critically. 
+Metrics marked as critical can also define `criticalIncrease` and `criticalDecrease`. 
+See the [Netflix Automated Canary Analysis Judge] and [Mann Whitney Classifier] classes for more information.
+
+#### Properties
+- `allowedIncrease` **1.1** (number, optional) - Defaults to 1. The multiplier increase that must be met for the metric to fail. This example makes the metric fail when the metric has increased 10% from the baseline.
+- `allowedDecrease` **0.90** (number, optional) - Defaults to 1. The multiplier decrease that must be met for the metric to fail. This example makes the metric fail when the metric has decreased 10% from the baseline.
+- `criticalIncrease` **5.0** (number, optional) - Defaults to 1. The multiplier increase that must be met for the metric to be a critical failure and fail the entire analysis with a score of 0. This example make the canary fail critically if there is a 5x increase.
```
fieldju: @csanden could you verify the accuracy of `criticalIncrease ` and `criticalDecrease` and there relationship with `critical` metrics.

I came up with this from reading the code, and i'm not a Scala expert.

docs/canary-config.md
fieldju: @nisanharamati, I dug through the Kayenta code and it doesn't have any of that logic. It only basically toggles generating an id or not, which is what I have documented for now. However we can probably add a note in the description about Orca if you want.

docs/canary-config.md
```diff
@@ -0,0 +1,100 @@
...
+<a name="links"></a>
+## Links
+- [Spinnaker Canary Best Practices]
+- [Canary analysis: Lessons learned and best practices from Google and Waze]
+- [Armory Kayenta Documentation]
+- [Example Signalfx canary config]
+  
+[Spinnaker Canary Best Practices]: https://www.spinnaker.io/guides/user/canary/best-practices/
+[Armory Kayenta Documentation]: https://docs.armory.io/spinnaker/configure_kayenta/
+[Example Signalfx canary config]: http:s//github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/metric-query-config.md
```
nisanharamati: typo in the url

nisanharamati:
```suggestion
[Example Signalfx canary config]: https//github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/metric-query-config.md
```

docs/canary-config.md
```diff
@@ -0,0 +1,100 @@
...
+<a name="links"></a>
+## Links
+- [Spinnaker Canary Best Practices]
+- [Canary analysis: Lessons learned and best practices from Google and Waze]
+- [Armory Kayenta Documentation]
+- [Example Signalfx canary config]
+  
+[Spinnaker Canary Best Practices]: https://www.spinnaker.io/guides/user/canary/best-practices/
+[Armory Kayenta Documentation]: https://docs.armory.io/spinnaker/configure_kayenta/
+[Example Signalfx canary config]: https//github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/metric-query-config.md
```
nisanharamati:
```suggestion
[Example Signalfx canary config]: https://github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/metric-query-config.md
```

nisanharamati: Sorry, I should have checked my own suggestion to be typo free!

fieldju: 🥇

docs/canary-config.md
```diff
@@ -0,0 +1,100 @@
...
+#### Properties
+- `name` **NetflixACAJudge-v1.0** (string, required) - Judge to use, as of right now there is only `NetflixACAJudge-v1.0`.
+- `judgeConfigurations` **{}** (object, required) - Map<string, object> of judgement configuration, this should always be an empty object as of right now.
+
+<a name="canary-metric-config"></a>
+### CanaryMetricConfig (object)
+Describes a metric that will be used in determining the health of a canary.
+#### Properties
+- `name` **http errors** (string, required) - Human readable name of the metric under test.
+- `query` (enum[[CanaryMetricSetQueryConfig](canary-metrics-set-query-config)], required) - Query config object for your metric source type.
```
nisanharamati:
```suggestion
- `query` (enum[[CanaryMetricSetQueryConfig](#canary-metrics-set-query-config)], required) - Query config object for your metric source type.
```

docs/canary-config.md
nisanharamati: I think it's okay to leave out in here. I'll try and think of where would be a good place to document that behaviour.

docs/canary-config.md
nisanharamati: IIRC when a metric goes over a critical threshold, the entire test is terminated to minimize negative production impact

csanden: That is correct. If a metric is marked a `critical` then we will check if the user has supplied a `criticalIncrease` or `criticalDecrease` value. These properties allow the user to control the sensitivity of a "critical" failure. Note that, a critical failure will result in the entire canary to be marked as a failure and receive a score of zero.

README.md
```diff
@@ -12,6 +12,9 @@ The quality of the canary version is assessed by comparing key metrics that desc
 Canaries are usually run against deployments containing changes to code, but they
 can also be used for operational changes, including changes to configuration.
 
+### Creating Canary Config
+Checkout the [Canary Config Object model](docs/canary-config.md) for how canary config is defined in [Markdown Syntax for Object Notation (MSON)](https://github.com/apiaryio/mson)
```
csanden:
```suggestion
Checkout the [Canary Config Object model](docs/canary-config.md) for how a canary config is defined in [Markdown Syntax for Object Notation (MSON)](https://github.com/apiaryio/mson)
```

PR #630, opened 2019-10-23
chore(docs): add base FAQ and Kayenta standalone docs.

fieldju:
Here is my first pass attempt at documenting everything that I have learned over the last year when it comes to running Kayenta as a standalone microservice.

This is my initial brain dump into MD files. A bunch of this might be better suited here: https://www.spinnaker.io/guides/user/canary/

This PR closes https://github.com/spinnaker/kayenta/issues/411

@nisanharamati it won't let me `@mention` you or let me add you as a reviewer.

Viewing the docs via this link might be useful: https://github.com/Nike-Inc/kayenta/blob/docs/kayenta_standalone/docs/faq.md


kayenta-signalfx/src/integration-test/resources/config/kayenta.yml
```diff
@@ -19,6 +19,7 @@ kayenta:
     - name: in-memory-store
       supportedTypes:
       - OBJECT_STORE
+      - CONFIGURATION_STORE
```
marchello2000: this and the other .yml don't seem to be part of docs

fieldju: Those where edits I made for [docs/configuring-kayenta.md](https://github.com/Nike-Inc/kayenta/blob/docs/kayenta_standalone/docs/configuring-kayenta.md) which is part of the PR.

In that doc I say that at least one configuration store is required but then link to a yml that didn't have one configured.

It's technically only required if you hit the `/config` endpoints which the integration test yml that I linked to does not.

The other yaml aka the big kayenta.yaml, I added a config option that was missing.
I think that yml isn't actually used and is more or less documentation itself.

fieldju:
Just a heads up @melanahammel is going to do a round of copy editing the docs in this PR and make a new commit with those changes. It might be worth your time to wait until this commit has been pushed before reviewing.

fieldju:
Im actually going to close the PR until that we have finished the initial rounds of copy editing.

PR #635, opened 2019-11-20
feat(retries): Add exponential backoff to query retries.

duftler:


duftler:
Would like to do the refactoring so everything (this spot, and there are several others) relies on resilience4j, but can't at the moment.

Was trying to keep this PR small to address the issue raised in https://github.com/spinnaker/kayenta/issues/634.

fieldju:
That makes sense, I was just having a shower thought for the future.

PR #637, opened 2019-11-26
chore(dependencies): Upgrade Spring Boot to 2.2.1

Jammy-Louie:
- See spinnaker/spinnaker#5134
- Dependent on spinnaker/kork#419 for a Spring Boot dependency upgrade


cfieber:
we should probably have merged keiko, then released that into orca, then updated orca into this as part of this change

PR #643, opened 2019-12-18
feat(*): Github Actions for PR builds

ajordens:


fieldju:
Would still like the functional tests to run in CI
https://github.com/spinnaker/kayenta/blob/master/.travis.yml#L14

ajordens:
Oh interesting ... that's non-standard but I'll add the same env variable.

fieldju:
Does this PR trigger the action or does it have to be merged and then it works with future PRs?
I was going to try to look at the logs and make sure all that docker stuff for the functional tests still works in the new CI env.

ajordens:
I think it'll have to merge first and then it'll run on future PRs.

This isn't taking away TravisCI builds (and we'll still be relying on a 👍 from it).

PR #645, opened 2019-12-20
chore: catch 403s and throw error when checking for GCS bucket

rarneson:
### What?
This PR includes the following changes to the `GcsStorageService` package:
- Check for a 403 response when checking if the bucket exists
- Throw a descriptive error on 403
- Link to the FAQ in the error message

The  FAQ was also updated to include an explanation of necessary permissions and a change Google made to their Standard Roles.

### Why?
The `GcsStorageService` checks for the existence of a bucket by making a GET request to the `storage/v1/b/{bucket}` API, which just returns metadata of the bucket if it exists. This API requires that your role has the `storage.buckets.get` permission. This permission used to be included in Google's Standard Roles, but for whatever reason has been moved into their Legacy Roles. You now have three options: add the legacy role `roles/storage.legacyBucketReader`, add the individual permission to a new, custom role, or use the `role/storage.admin` role.


kayenta-gcs/src/main/java/com/netflix/kayenta/gcs/storage/GcsStorageService.java
```diff
@@ -100,6 +100,12 @@ public void ensureBucketExists(String accountName) {
           log.error("Could not create bucket {} in project {}: {}", bucketName, projectName, e2);
           throw new IllegalArgumentException(e2);
         }
+      } else if (e.getStatusCode() == 403) {
+        log.error(
+            "Account does not have permissions to get bucket metadata {}. Please see the FAQ for details: https://github.com/spinnaker/kayenta/blob/master/docs/faq.md: {}",
```
duftler: Looks like you can deep-link directly to the new faq entry (once it's merged). Likely just need to add `#my-gcs-account-does-not-have-access-to-get-bucket-metadata`.

docs/faq.md
```diff
@@ -54,3 +54,7 @@ _Warning: This is a beta feature and may be removed._
 
 Yes, there are a couple of settings available for you.
 See [EffectSize](./canary-config.md#effect-size) for more information.
+
+## My GCS account does not have access to get bucket metadata.
+
+In order for Kayenta to read and write from a GCS bucket, it first needs to checks for the existence of the bucket. It does this by making a GET request to the `storage/v1/b/{bucket}` API which returns some metadata about the bucket. In order to interact with this API you need a role that has the `storage.buckets.get` permission. This permission used to be included in Google's [Standard Roles](https://cloud.google.com/storage/docs/access-control/iam-roles), but has since been removed and put into [Legacy Roles](https://cloud.google.com/storage/docs/access-control/iam-roles#legacy-roles). In order to get that permission, you can either create a custom role and apply `storage.buckets.get` to it, or you can add the `roles/storage.legacyBucketReader` as explained in the Legacy Roles section.
```
duftler: The `roles/storage.admin` role includes permission `storage.buckets.*`: 
image

kayenta-gcs/src/main/java/com/netflix/kayenta/gcs/storage/GcsStorageService.java
rarneson: Good catch.

docs/faq.md
rarneson: I'll make note of that as another option.

PR #648, opened 2020-01-06
chore(dependencies): Reverting Spring Boot 2.2 upgrade

pdelagrave:
Revert "chore(dependencies): Upgrade Spring Boot to 2.2.1 (#637)"
This reverts commit e2b1d7b4


fieldju:
The build seems to be broken, also why does this need to be reverted?

duftler:
Context for the revert(s) can be found on this kork pr: https://github.com/spinnaker/kork/pull/459

duftler:
Build should be green once the orca version this PR depends on is published. Working through that now.

PR #657, opened 2020-01-27
feat(datadog): add support for inline template

anthonywoo:
Add inline template support for Datadog to allow for custom queries instead of a single metric

Example canary config with inline template:
```
"query": {
  "type": "datadog",
  "serviceType": "datadog",
  "customInlineTemplate":"avg:app.errors{${scope}}/avg:app.requests{${scope}}"
},
```

I've tested this with the kayenta standalone setup and validated it with running `/canary` and `/standalone_canary_analysis`


fieldju:
These change looks pretty reasonable.

I would love to see unit test similar to:

https://github.com/spinnaker/kayenta/blob/master/kayenta-prometheus/src/test/groovy/com/netflix/kayenta/prometheus/metrics/PrometheusMetricsServiceSpec.groovy

https://github.com/spinnaker/kayenta/blob/master/kayenta-newrelic-insights/src/test/groovy/com/netflix/kayenta/newrelic/metrics/NewRelicQueryBuilderServiceSpec.groovy

https://github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/src/test/java/com/netflix/kayenta/signalfx/metrics/SignalFxQueryBuilderServiceTest.java

https://github.com/spinnaker/kayenta/blob/master/kayenta-signalfx/src/test/java/com/netflix/kayenta/signalfx/metrics/SignalFxQueryBuilderServiceTest.java

@duftler / @nisanharamati should unit tests be mandatory?

anthonywoo:
Thanks for pointing me to those example tests. I've added a similar spec to cover the inline template use case.

kayenta-datadog/src/test/groovy/com/netflix/kayenta/datadog/metrics/DatadogMetricsServiceSpec.groovy
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
fieldju: This is supposed to be your company / you. Unless you do work for Netflix and I missed that.

anthonywoo: got it. Changed company

duftler:
> @duftler / @nisanharamati should unit tests be mandatory?

I don't think we need to go so far as to mandate them, but seems reasonable to expect them when there are similar test examples.

PR #659, opened 2020-01-29
Replace freemarker with stringsubstitutor from commons-text for template expanding

Aloren:
Freemarker is very powerful template engine, but in Kayenta only 1 percent of it's capabilities are actually used. For template expanding with map of variables simple string substitutor does the trick.
Also this change allowed to remove `throws IOException` from the signature of `MetricsService.buildQuery`.

Added tests for all basic scenarios of `QueryConfigUtils`.

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
@@ -21,15 +21,10 @@ import com.netflix.kayenta.canary.CanaryConfig;
 import com.netflix.kayenta.canary.CanaryMetricConfig;
 import com.netflix.kayenta.canary.CanaryMetricSetQueryConfig;
 import com.netflix.kayenta.canary.CanaryScope;
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
 import java.beans.BeanInfo;
 import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
-import java.io.IOException;
-import java.io.StringReader;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collections;
 import java.util.LinkedHashMap;
@@ -39,7 +34,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.ui.freemarker.FreeMarkerTemplateUtils;
+import org.apache.commons.text.StringSubstitutor;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -51,8 +46,7 @@ public class QueryConfigUtils {
       CanaryConfig canaryConfig,
       CanaryMetricSetQueryConfig metricSetQuery,
       CanaryScope canaryScope,
-      String[] baseScopeAttributes)
-      throws IOException {
+      String[] baseScopeAttributes) {
     if (metricSetQuery.getCustomFilter() != null) {
       throw new IllegalArgumentException(
           "CanaryMetricSetQueryConfig.customFilter is deprecated, use CanaryMetricSetQueryConfig.customInlineTemplate instead.");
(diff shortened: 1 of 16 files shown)


Aloren:
Also regarding the issue with the newly added test `expandCustomFilter_returnsExpandedTemplateWithExtractedBaseScopeAttributesFromFilterTemplate`. If baseScopeAttribute contains properties both from `CanaryConfig` and `CanaryMetricSetQueryConfig` ( like "scope", "location", "serviceType" ), `populateTemplateBindings` will fail with exception `java.lang.IllegalArgumentException: Unable to find property 'serviceType'.` Looks like this is a bug in `populateTemplateBindings`. If you don't mind I would like to address that issue in a separate pull request.

kayenta-stackdriver/src/test/groovy/com/netflix/kayenta/providers/metrics/StackdriverCanaryMetricSetQueryConfigSpec.groovy
```diff
@@ -92,7 +92,6 @@ class StackdriverCanaryMetricSetQueryConfigSpec extends Specification {
     ["my-template-1": 'A test: key1=${key1}.',
      "my-template-2": 'A test: key2=${key2}.']                | "my-template-x"        | null
     [:]                                                       | "my-template-x"        | null
-    null                                                      | "my-template-x"        | null
```
Aloren: Added `@Singular` for `templates` field in `CanaryConfig`, so it is not valid to use: `CanaryConfig.builder().templates(null).build()`, that's why I deleted this case.

fieldju:
I am going to defer review of this to @duftler as he made the original template integration and could speak to this change better than I.

Out of curiosity what is the problem this PR solves?

Aloren:
I was working on controller tests and saw that the context was starting too slow, digged into the context and found freemarker in the dependencies. Initially thought that this was added by mistake, but then found that this was used for replacing placeholders in string. I personally think that this is excessive and since the change was trivial I submitted PR. (Also it is possible to use standard spel for such things as an alternative)

kayenta-core/src/test/groovy/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtilsTest.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
duftler:
```suggestion
 * Copyright 2020 Playtika, Inc.
```

kayenta-core/src/test/groovy/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtilsSpec.groovy
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright 2018 Google, Inc.
+ * Copyright 2020 Netflix, Inc.
```
duftler:
```suggestion
 * Copyright 2018 Google, Inc.
```

kayenta-core/src/test/groovy/com/netflix/kayenta/canary/providers/metrics/TestCanaryMetricSetQueryConfig.groovy
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
duftler:
```suggestion
 * Copyright 2020 Playtika, Inc.
```

duftler:
One other minor thing, I think the commit message should be: "refactor(core): Replace freemarker..." instead of "feat(core): Replace freemarker...".

kayenta-core/src/main/java/com/netflix/kayenta/canary/providers/metrics/QueryConfigUtils.java
```diff
@@ -51,8 +46,7 @@ public static String expandCustomFilter(
       CanaryConfig canaryConfig,
       CanaryMetricSetQueryConfig metricSetQuery,
       CanaryScope canaryScope,
-      String[] baseScopeAttributes)
-      throws IOException {
```
nisanharamati: Nice!

Aloren:
Absolutely love the part when you need to fix copyright headers.. :D

Aloren:
> Once the copyright messages are fixed and the commit message is updated, all that remains is for someone from Netflix to have a look.

Just fixed.
Thank you for the review!

Issue #354, opened 2018-07-20
Provide more meaningful error messages in response to kayenta canary context parsing errors


duftler:
When the orca logs show:
```
2018-07-20 17:36:14.651 ERROR 1 --- [    handlers-18] c.n.s.orca.q.handler.StartStageHandler   : [] Error running kayentaCanary stage for pipeline[01CJWC57PJKCB26FN5RP54584Y]
java.lang.IllegalArgumentException: Unable to map context to class com.netflix.spinnaker.orca.kayenta.model.KayentaCanaryContext
	at com.netflix.spinnaker.orca.pipeline.model.Stage.mapTo(Stage.java:434)
	at com.netflix.spinnaker.orca.kayenta.pipeline.KayentaCanaryStage.beforeStages(KayentaCanaryStage.kt:199)
	at com.netflix.spinnaker.orca.q.StageDefinitionBuildersKt.buildBeforeStages(StageDefinitionBuilders.kt:83)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler.plan(StartStageHandler.kt:141)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler.access$plan(StartStageHandler.kt:48)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler$handle$1$1.invoke(StartStageHandler.kt:81)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler$handle$1$1.invoke(StartStageHandler.kt:48)
	at com.netflix.spinnaker.orca.q.handler.AuthenticationAware$sam$java_util_concurrent_Callable$0.call(AuthenticationAware.kt)
	at com.netflix.spinnaker.security.AuthenticatedRequest.lambda$propagate$0(AuthenticatedRequest.java:60)
	at com.netflix.spinnaker.orca.q.handler.AuthenticationAware$DefaultImpls.withAuth(AuthenticationAware.kt:49)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler.withAuth(StartStageHandler.kt:48)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler$handle$1.invoke(StartStageHandler.kt:80)
	at com.netflix.spinnaker.orca.q.handler.StartStageHandler$handle$1.invoke(StartStageHandler.kt:48)
	at com.netflix.spinnaker.orca.q.handler.OrcaMessageHandler$withStage$1.invoke(OrcaMessageHandler.kt:58)
[...]

PR #662, opened 2020-01-31
fix(web): Enhance exception handling to return detailed message for the standard exceptions

Aloren:
- Enhance exception handling to return detailed message for the standard exceptions (like invalid requests) handled by Spring (see: ResponseEntityExceptionHandler). 
- Enhance controller method definitions.

Motivation behind this change: when user sends invalid request or any other internal error occurs, he/she gets only http status in response and no reason for what is the issue; this pull request addresses this issue and adds simple json response body with `message` field.

### DIFF
--- a/kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java
+++ b/kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java
@@ -19,13 +19,17 @@ package com.netflix.kayenta.controllers;
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import java.util.Collections;
 import java.util.Map;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 
 @RestControllerAdvice
-public class ControllerExceptionHandler {
+public class ControllerExceptionHandler extends ResponseEntityExceptionHandler {
 
   @ResponseStatus(HttpStatus.BAD_REQUEST)
   @ExceptionHandler(IllegalArgumentException.class)
@@ -37,6 +41,12 @@ public class ControllerExceptionHandler {
   @ExceptionHandler(NotFoundException.class)
   public Map handleException(NotFoundException e) {
     return toErrorResponse(e);
+  }
+
+  @Override
+  protected ResponseEntity<Object> handleExceptionInternal(
+      Exception ex, Object body, HttpHeaders headers, HttpStatus status, WebRequest request) {
+    return super.handleExceptionInternal(ex, toErrorResponse(ex), headers, status, request);
   }
 
   private Map toErrorResponse(Exception e) {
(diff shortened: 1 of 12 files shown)


kayenta-web/src/main/java/com/netflix/kayenta/controllers/ControllerExceptionHandler.java
```diff
@@ -19,13 +19,17 @@
...
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 
 @RestControllerAdvice
-public class ControllerExceptionHandler {
+public class ControllerExceptionHandler extends ResponseEntityExceptionHandler {
```
Aloren: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/mvc/method/annotation/ResponseEntityExceptionHandler.html

fieldju:
I am all about more context in the returned errors: see https://github.com/spinnaker/kayenta/issues/354

Are the changes you are making backwards compatible with Kayentas Deck and Orca integrations?

What are the before and after shapes of the error responses?

fieldju:
Does this sort of logic belong in https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java?

Im not 100% sure what the pattern is supposed to be with Kork vs implementing this in the individual Spinnaker services. It seems like the exception handling logic should be consistent across the whole Spinnaker ecosystem.

Maybe @cfieber or @robzienert could chim in here on Spinnaker best practices?

Aloren:
> Does this sort of logic belong in https://github.com/spinnaker/kork/blob/master/kork-web/src/main/java/com/netflix/spinnaker/kork/web/exceptions/GenericExceptionHandlers.java?

Thank you. I was looking initially for such sort of thing, but somehow missed it. I'm not sure, but it looks to me that this is actually not used in Kayenta. Will check that now.

Aloren:
Okay, so had a look into /beans endpoint and found:
```
"genericExceptionHandlers": {
"aliases": [],
"scope": "singleton",
"type": "com.netflix.spinnaker.kork.web.exceptions.GenericExceptionHandlers",
"resource": "class path resource [com/netflix/spinnaker/config/ErrorConfiguration.class]",
"dependencies": []
},
```
so this is actually setup.

I commented out the newly added behavior in `ControllerExceptionHandler` and added `ErrorConfiguration` into the tests setup. Now 
`storeCanaryConfig_returnsBadRequestResponseForMissingRequestBody` fails with:
```
MockHttpServletResponse:
           Status = 500
    Error message = Required request body is missing: public com.netflix.kayenta.canary.CanaryConfigUpdateResponse com.netflix.kayenta.controllers.CanaryConfigController.storeCanaryConfig(java.lang.String,com.netflix.kayenta.canary.CanaryConfig)
          Headers = [X-Content-Type-Options:"nosniff", X-XSS-Protection:"1; mode=block", Cache-Control:"no-cache, no-store, max-age=0, must-revalidate", Pragma:"no-cache", Expires:"0", X-Frame-Options:"DENY"]
     Content type = null
             Body = 
    Forwarded URL = null
   Redirected URL = null
          Cookies = []
```

Aloren:
I suppose that `extends ResponseEntityExceptionHandler` should be added into `GenericExceptionHandlers` (`kork-web`), so that all spring-mvc exceptions were processed correctly across all spinnaker web applications.

robzienert:
1. Thank you for cleaning up the `RequestMapping` annots.
1. I'd prefer to see improvements around our RPC exceptions to be applied across all projects. So, I'd prefer to see those changes land in kork if at all possible.

robzienert:
Ping. I like the contribution but I think it should be more broadly applied across the Spinnaker services.

Aloren:
@robzienert I will move implementatio to kork-web module then and leave here only endpoints enhancements. Thanks

Aloren:
closing in favor of: 
https://github.com/spinnaker/kayenta/pull/681
https://github.com/spinnaker/kork/pull/535

PR #668, opened 2020-02-04
feat(core): move CanaryJudgesController to kayenta-web module

Aloren:
CanaryJudgesController is the only controller that is located in kayenta-core module. All other Kayenta web endpoints are located in kayenta-web.


kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryJudgesController.java
```diff
@@ -1,7 +1,7 @@
 /*
- * Copyright 2017 Google, Inc.
+ * Copyright 2020 Netflix, Inc.
```
duftler:
```suggestion
 * Copyright 2020 Google, Inc.
```

PR #421, opened 2018-11-09
Allow metrics to be specified as must have data.

fieldju:
While doing some testing I discovered that I wanted the feature described in https://github.com/spinnaker/kayenta/issues/330.

This PR is an attempt at solving that request, hopefully this is something close to what you were thinking about.

PR #676, opened 2020-02-26
refactor(judge): remove isDataRequired field from MetricClassification

Aloren:
This field was introduced in #421, but it is never set and never used anywhere.


fieldju:
https://github.com/Nike-Inc/kayenta/blob/8c700e66fdec052e0c29345510031773cf538636/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/MannWhitneyClassifier.scala#L139

fieldju:
This field is used to determine when a metric should fail because it has no data and data is required.

fieldju:
I was having issues with my web browser, this should be fine.

fieldju:
Will merge once the build finishes, had to merge master in.

csanden:
Just to clarify for future reference, the `isDataRequired` field is not used as part of the `MetricClassification` result, i.e., what the metric classifiers return. However, there is a `isDataRequired` parameter as part of the  `classify` method on the `BaseMetricClassifier` which is still used.

PR #681, opened 2020-03-02
refactor(web): Enhance controller method definitions

Aloren:


kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
```diff
@@ -190,11 +185,11 @@ private static void validateNameAndApplicationAttributes(@RequestBody CanaryConf
   }
 
   @ApiOperation(value = "Delete a canary config")
-  @RequestMapping(value = "/{canaryConfigId:.+}", method = RequestMethod.DELETE)
+  @ResponseStatus(HttpStatus.NO_CONTENT)
```
fieldju: Is this a breaking change?

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryResultArchiveController.java
```diff
@@ -134,22 +132,20 @@ public CanaryArchiveResultUpdateResponse updateArchivedCanaryResult(
   }
 
   @ApiOperation(value = "Delete an archived canary result from object storage")
-  @RequestMapping(value = "/{pipelineId:.+}", method = RequestMethod.DELETE)
+  @DeleteMapping(value = "/{pipelineId:.+}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
```
fieldju: Same question here as above.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSetListController.java
```diff
@@ -85,24 +86,22 @@ public Map storeMetricSetList(
   }
 
   @ApiOperation(value = "Delete a metric set list")
-  @RequestMapping(value = "/{metricSetListId:.+}", method = RequestMethod.DELETE)
+  @DeleteMapping(value = "/{metricSetListId:.+}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
```
fieldju: Same question here as above.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricSetPairListController.java
```diff
@@ -113,11 +112,11 @@ public Map storeMetricSetPairList(
   }
 
   @ApiOperation(value = "Delete a metric set pair list")
-  @RequestMapping(value = "/{metricSetPairListId:.+}", method = RequestMethod.DELETE)
+  @DeleteMapping(value = "/{metricSetPairListId:.+}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
```
fieldju: Same question here as above.

fieldju:
@Aloren I see a few technically breaking changes in the API contract as part of this PR.
You appear to be changing the response code for delete calls.

1. Does this change align with the behavior in the rest of the Spinnaker APIs?
2. Have you tested this change against Deck/Gate to ensure this breaking change doesn't break the existing user experiences in the Spinnaker system.

kayenta-web/src/main/java/com/netflix/kayenta/controllers/CanaryConfigController.java
Aloren: no, this was already there before. I extracted this to the annotation only

Aloren:
Sorry for the delayed answer. This PR does not change the contract for any method.
Previously https status was set via:
```
response.setStatus(HttpStatus.NO_CONTENT.value());
```
I replaced it with 
```
@ResponseStatus(HttpStatus.NO_CONTENT)
```
So this is still the same logic.

kayenta-web/src/test/java/com/netflix/kayenta/controllers/CanaryConfigControllerTest.java
```diff
@@ -63,9 +63,30 @@ public void getCanaryConfig_returnsBadRequestResponseForNotResolvedAccount() thr
...
   }
+
+  @Test
+  public void storeCanaryConfig_returnsBadRequestResponseForMissingRequestBody() throws Exception {
+    this.mockMvc
+        .perform(
+            post("/canaryConfig?configurationAccountName={account}", CONFIGS_ACCOUNT)
+                .contentType(APPLICATION_JSON))
+        .andExpect(status().isInternalServerError()); // expected: bad request ( should be fixed by:
+    // https://github.com/spinnaker/kork/pull/535)
```
fieldju, line 78: This test is still failing even though that is merged, I think orca will need to be bumped so that Kork is updated

- [ ] tests pass

fieldju:
The build is failing on this PR, therefor I cannot merge it.
I am adding the stale label on this and will close the PR in 30 days.

PR #687, opened 2020-03-19
chore(build): kotlin 1.3.70

sullis:


sullis:
The build failed due to a problem with a Scala source file:

```
kayenta/kayenta-judge/src/main/scala/com/netflix/kayenta/judge/classifiers/metric/MannWhitneyClassifier.scala:38: not found: type MannWhitney

    val mw = new MannWhitney()

                 ^
```

This error is not related to my change.

sullis:
obsolete

PR #690, opened 2020-03-23
refactor(prometheus): extract metricDescriptorsCache to PrometheusMetricDescriptorsCache

Aloren:
- enhance(prometheus): extract metricDescriptorsCache to PrometheusMetricDescriptorsCache
- fix(web): remove erroneous second call to getMetadata in MetricsServiceMetadataController


kayenta-orca/src/main/java/com/netflix/kayenta/orca/controllers/PipelineController.java
```diff
@@ -87,7 +87,7 @@ void startOrcaQueueProcessing() {
         postProcessor.postProcessBeforeDestruction(this, null);
         log.info("Health indicators are all reporting UP; starting orca queue processing");
       } else {
-        log.info(
+        log.warn(
```
Aloren: I think more appropriate level for this log is warning.

kayenta-prometheus/src/main/java/com/netflix/kayenta/prometheus/metrics/PrometheusMetricDescriptorsCache.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Slf4j
+public class PrometheusMetricDescriptorsCache {
+
+    private volatile Map<String, List<PrometheusMetricDescriptor>> cache = Collections.emptyMap();
```
Aloren: this was not volatile before, but it can be updated in the different thread, so we do want to guarantee that all threads receive the changes.

kayenta-prometheus/src/main/java/com/netflix/kayenta/prometheus/metrics/PrometheusMetricDescriptorsCache.java
```diff
@@ -0,0 +1,131 @@
...
+                    .map(metricDescriptor -> metricDescriptor.getMap())
+                    .collect(Collectors.toList());
+        }
+    }
+
+    @Scheduled(fixedDelayString = "#{@prometheusConfigurationProperties.metadataCachingIntervalMS}")
+    public void updateMetricDescriptorsCache() {
+        Set<AccountCredentials> accountCredentialsSet =
+                accountCredentialsRepository.getAllOf(AccountCredentials.Type.METRICS_STORE);
+
```
Aloren: replaced existing code with the streams api without any changes in logic. this way it looks more simple

kayenta-prometheus/src/main/java/com/netflix/kayenta/prometheus/metrics/PrometheusMetricDescriptorsCache.java
```diff
@@ -0,0 +1,131 @@
...
+                    "While updating cache, found no metric descriptors via account {}.", accountResponse.getMetricsAccountName());
+        }
+        return !empty;
+    }
+
+    private boolean isSuccessful(AccountResponse accountResponse) {
+        PrometheusMetricDescriptorsResponse response = accountResponse.getResponse();
+        boolean success = response != null && response.getStatus().equals("success");
+        if (!success) {
+            log.debug("While updating cache, found no metric descriptors via account {}.", accountResponse.getMetricsAccountName());
```
Aloren: this was the original message in this code, but I suppose it should be something different, because the request actually failed. please let me know if I need to change it or leave it as is

kayenta-web/src/main/java/com/netflix/kayenta/controllers/MetricsServiceMetadataController.java
```diff
@@ -77,6 +76,6 @@ public MetricsServiceMetadataController(
           filter);
     }
 
-    return metricsService.getMetadata(resolvedMetricsAccountName, filter);
+    return matchingDescriptors;
```
Aloren: there was an erroneous second call to `getMetadata`, removed it

PR #697, opened 2020-04-02
fix(plugins): add an option for strict plugin loading

link108:
This PR is part of a series of PRs for allowing relaxed plugin loading by adding a config option to turn strict plugin loading on or off. See https://docs.google.com/document/d/1yFDrpkCCV7Vp0wN0gDc9GTbTOBoIzNo-tGj1-vVjq04/edit?usp=sharing for more info

PR #698, opened 2020-04-03
fix(plugins): add an option for strict plugin loading (#697)

link108:
Similar to https://github.com/spinnaker/kayenta/pull/697, mergify's backport failed


fieldju:
@ezimanyi would a reasonable alternative for 1.19.x be to revert Orca to the latest 7.x and just manually pull in the version of Kork that is wanted. I am naively assuming kork 7.30.1 will work with Orca 7.+ 

Something like

```groovy
configurations.all {
  resolutionStrategy {  
    eachDependency { DependencyResolveDetails details ->
      if (details.requested.group == 'com.netflix.spinnaker.kork') {
        details.useVersion "7.30.1"
      }
    }
  }
}
```

fieldju:
image

Forcing kork `7.25.2` -> `7.25.13` to be inline with the reset of the `1.19.x` releases and allow for releaxed plugin loading.

ezimanyi:
Do we even need to bump `orca` at all?  Even though we're not doing a major version bump, we're still pulling in a lot of [commits](https://github.com/spinnaker/orca/compare/v7.78.0...v7.87.0), many of which are refactors and features that we wouldn't want in a release branch.  Is there a reason we can't just pull the new `kork` in and leave `orca` as is?

fieldju:
> Do we even need to bump `orca` at all? Even though we're not doing a major version bump, we're still pulling in a lot of [commits](https://github.com/spinnaker/orca/compare/v7.78.0...v7.87.0), many of which are refactors and features that we wouldn't want in a release branch. Is there a reason we can't just pull the new `kork` in and leave `orca` as is?

@ezimanyi Let me double check, when I tried using the latest kork, the integration tests failed. I have since down graded kork to `7.25.13` which is what the rest of the 1.19.x branches all use. 

I will revert the orca change and run the integration tests 1 more time and see what the results are.

fieldju:
image

@ezimanyi, somewhere between Orca 7.78 and 7.87 something has changed with Kork, because the stages for the canary runs never start.

I can attempt to find the min required bump for Orca if we want.

ezimanyi:
I guess I don't feel overly confident cherry-picking this in if something changed in the interaction between kork, orca, and kayenta across these versions and we're not sure what.

The safest option is probably:
* Find the `kork` version in `orca 7.78.0`, make a branch from that commit of `kork` and cut a new `kork` release
* Make `orca 7.78.1` which just pulls in that new `kork` version
* Pull `orca 7.78.1` here

It's unfortunately a lot of steps but if we don't do that, we're effectively pulling in a bunch of other changes into this cherry-pick which risks breaking people on a stable release branch.

fieldju:
@ezimanyi 

Kayenta 1.19.x is using kork 7.13.0, kork has a 7.13.2 already so I assume we should cherry pick the plugin fix into 7.13.2 and release as 7.13.3

@link108 Can you checkout the `7.13.2` tag of Kork and cherry-pick your plugin changes.

@ezimanyi Where should we PR this to, in order to get peer review before making the 7.13.3 tag which will trigger the bintray release.

@ezimanyi also we can skip the last 2  bullets if we do this in kayenta

```
configurations.all {
  resolutionStrategy {  
    eachDependency { DependencyResolveDetails details ->
      if (details.requested.group == 'com.netflix.spinnaker.kork') {
        details.useVersion "7.13.3"
      }
    }
  }
}
```

ezimanyi:
As I look at it more, I'm not sure that's even going to work...it looks like there has been a lot of plugin work on `kork` since `7.13.0` so it's not clear this will cherry pick cleanly.  I'll defer to @link108 to see if this can be cherry-picked there.

If the cherry-pick does work then I can make a new branch in `spinnaker/kork` to accept the pull request and from which to cut `7.13.3`.  But please test locally first, and I can create the branch once we know this will work.

And good point, it may just be easier to skip the `orca` part and just pin kork to `7.13.3` if this works.

fieldju:
i'd almost just ask if Kayenta can just wait till `1.20.0` @link108 ?

PR #699, opened 2020-04-08
fix(standalone-canary-analysis): custom parent pipeline execution id is not used

Aloren:
When I retrieve results for the standalone canary analysis via archive endpoint I see that `parentPipelineExecutionId` is already set, but I haven't provided it. This PR fixes `parentPipelineExecutionId` to be equal to the one that was set by the user.

@fieldju could you please review?


fieldju:
This looks good, sorry I have been on paternity leave. I am back now.

PR #708, opened 2020-04-19
chore(build): update to newer spinnaker project plugin

cfieber:
updates to gradle 6.3
updates to kotlin 1.3.70
updates github actions workflows for new release flags

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,6 @@
-enablePublishing=false
 keikoVersion=3.4.0
-kotlinVersion=1.3.20
+kotlinVersion=1.3.70
 mathCommonsVersion=3.6.1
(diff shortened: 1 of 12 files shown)


gradle.properties
```diff
@@ -1,7 +1,6 @@
-enablePublishing=false
 keikoVersion=3.4.0
-kotlinVersion=1.3.20
+kotlinVersion=1.3.70
```
jonsie, line 2: We've been running 1.3.71 in other repos.

cfieber: I'm going to leave it as is for now and figure out a way to include pushing the kork-bom kotlinVersion to the kork consumers as part of autobumping

PR #714, opened 2020-04-28
feat(plugins): plugin support for kayenta

link108:
This PR makes Kayenta pluggable, similar to spinnaker/clouddriver#4383


kayenta-core/kayenta-core.gradle
```diff
@@ -2,6 +2,7 @@ dependencies {
   api "redis.clients:jedis"
   api "com.netflix.spinnaker.kork:kork-core"
   api "com.netflix.spinnaker.kork:kork-web"
+  api "com.netflix.spinnaker.kork:kork-plugins"
```
danielpeach, line 5: Think this should be `implementation` under `kayenta-web` until we write an ExtensionPoint for kayenta.

PR #740, opened 2020-06-04
refactor(core): make ObjectType an interface instead of enum

Aloren:
make ObjectType an interface instead of enum to be able to use StorageService for some custom object types


fieldju:
I need to refresh my memory on a couple things.
But I have been wanting this for a bit: https://github.com/spinnaker/kayenta/pull/593

In that PR, I had to move all the models for standalone canary analysis into core, because of this enum.

Does this change enable me to move those models out of core and back into the standalone canary analysis sub project?

Aloren:
I had to do this change because it is impossible to use StorageService for custom data like report that is not present in the original ObjectType enum. Now since ObjectType is an interface, each module can define it's own ObjectType without the need to store domain models in core project! :) Do you want me to move standalone canary analysis domain models into the standalone canary analysis module? :)

fieldju:
No, you don't have to do that as part of this PR. If you wanted to do that in another PR that would be cool, or else maybe I can get to it eventually.

PR #741, opened 2020-06-05
refactor(web): extract logic from MetricSetPairListController into MetricSetPairListService to make this code reusable

Aloren:


kayenta-core/src/main/java/com/netflix/kayenta/service/MetricSetPairListService.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
duftler, line 2:
```suggestion
 * Copyright 2020 Playtika, Inc.
```

PR #742, opened 2020-06-10
refactor(core): move standalone canary analysis domain models back into it's module

Aloren:
Improvement that was discussed in https://github.com/spinnaker/kayenta/pull/740

@fieldju is moving these domain models to the different package considered safe? I've looked into storage service, looks like these are saved using json, but probably I've missed something...


fieldju:
It should be reasonably safe, it will break import statements on anyone who is extending the project with a wrapper such as @melanahammel.

But there is precedence of this being done without major versioning projects already within Spinnaker.

PR #747, opened 2020-06-20
chore(build): enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


fieldju:
I'm not going to merge this because there is already other dep management stuff going on across the Spinnaker projects and would want to see something like this being applied or merged broadly and not just in this Project.

PR #762, opened 2020-07-10
feat(judge): Add new Effect Size measure (CLES)

csanden:
Add the [Common Language Effect Size (CLES)](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test#Common_language_effect_size) as an alternative way to compute the effect size measure.  

The CLES reports the probability that a value from one group will be greater than a value from the other group.  A value of 0.50 indicates that the two groups are stochastically equal.  A value of 1 indicates that the first group shows complete stochastic domination over the other group, and a value of 0 indicates the complete stochastic domination by the second group.


vuxuanlai:
@nisanharamati , could you plz have an example about using CLES in canary config?

PR #766, opened 2020-07-21
config(kayenta-core): Add KayentaSerializationConfigurationProperties

nisanharamati:
A recent change to the `springboot.jackson.ObjectMapper`'s serialization behaviour changed the way `canaryDuration` is serialized in canary reports. This PR adds a configuration section to control that behaviour, and sets to default to match the behaviour before the change.

Jackson 2.10+ changes the default behaviour for serializing Duration from String to Timestamp.
The behaviour is already being set explicitly for Date objects, but wasn't addressed for Duration objects.
Rather than setting this property explicitly, I added a configuration stanza to `kayenta.yml` to allow users to set their preferred behaviour, and defaulting to the old behaviour of encoding both Date and Duration as Strings.

Before (with the updated serialization behaviour):
```
"canaryDuration": 1800000
```

After (with `writeDurationsAsTimestamps: false` in `kayenta.yml`):
```
"canaryDuration": "PT30M"
```

You can set either `Date`, `Duration`, or both to be serialized as Integer milliseconds by setting
```
kayenta:
  serialization:
    writeDatesAsTimestamp: true
    writeDurationsAsTimestamp: true
```
in `kayenta.yml`

### DIFF
--- a/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google, Inc.
+ * Copyright 2020 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -16,12 +16,12 @@
 
 package com.netflix.kayenta.atlas.config;
 
-import lombok.Getter;
-import lombok.Setter;
+import lombok.Data;
 
+@Data
 public class KayentaSerializationConfigurationProperties {
 
-  @Getter @Setter private boolean writeDatesAsTimestamps = false;
+  private boolean writeDatesAsTimestamps = false;
 
-  @Getter @Setter private boolean writeDurationsAsTimestamps = false;
+  private boolean writeDurationsAsTimestamps = false;
 }
--- a/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
+++ b/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.atlas.config;
+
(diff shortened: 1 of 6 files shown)


kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2017 Google, Inc.
```
duftler, line 2: s/2017 Google/2020 Netflix

kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaSerializationConfigurationProperties.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.netflix.kayenta.atlas.config;
+
+import lombok.Getter;
+import lombok.Setter;
+
+public class KayentaSerializationConfigurationProperties {
+
+  @Getter @Setter private boolean writeDatesAsTimestamps = false;
```
duftler, line 24: Why not just use `@Data` on the class and drop all the `@Getters` and `@Setters`?

kayenta-core/src/test/groovy/com/netflix/kayenta/metrics/CanaryScopeSpec.groovy
```diff
@@ -58,7 +57,9 @@ class CanaryScopeSpec extends Specification {
 
   private ObjectMapper myObjectMapper() {
     ObjectMapper objectMapper = new ObjectMapper();
-    KayentaConfiguration.configureObjectMapperFeatures(objectMapper);
+    KayentaSerializationConfigurationProperties kayentaSerializationConfigurationProperties = new KayentaSerializationConfigurationProperties();
+    kayentaSerializationConfigurationProperties.setWriteDatesAsTimestamps(false);
```
duftler, line 61: It looks like the property defaults to `false`, so does this setting have any effect?

Does the test fail if this property is set to `true`?

duftler:
I can't see how the existing tests verify that the two fields are in fact now serialized as a strings. If the goal is to prevent the public-facing contract from changing unexpectedly, I don't think the tests actually do that.

The `CanaryScopeSpec` tests pass whether the 2 new properties are set to `true` or `false`. This is because the  "should render as json and come back again" test makes an assertion over the round-tripped `CanaryScope` pojo. So what we are verifying is that we can deserialize the json no matter whether the dates/durations are serialized as strings or as timestamps.

I think `CanaryScopeSpec` could use an additional test verifying that dates/durations are serialized based on the values of the 2 new properties. Once that test is in place, we are no longer susceptible to the external behavior changing without us knowing about it.

nisanharamati:
> I can't see how the existing tests verify that the two fields are in fact now serialized as strings. If the goal is to prevent the public-facing contract from changing unexpectedly, I don't think the tests actually do that.
> 
> The `CanaryScopeSpec` tests pass whether the 2 new properties are set to `true` or `false`. This is because the "should render as json and come back again" test makes an assertion over the round-tripped `CanaryScope` pojo. So what we are verifying is that we can deserialize the json no matter whether the dates/durations are serialized as strings or as timestamps.
> 
> I think `CanaryScopeSpec` could use an additional test verifying that dates/durations are serialized based on the values of the 2 new properties. Once that test is in place, we are no longer susceptible to the external behavior changing without us knowing about it.

I agree. I'll make it its own test outside of `CanaryScopeTest`, as I don't think the serialization configuration is strictly tied to the `CanaryScope`

PR #768, opened 2020-07-21
chore(build): enable Gradle wrapper validation (GitHub Actions)

sullis:
https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Gradle validation step
+        uses: gradle/wrapper-validation-action@v1
       # Install Java 8 for cross-compilation support. Setting it up before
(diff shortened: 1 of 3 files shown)


sullis:
@fieldju

.github/workflows/build.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1
```
jan25, line 19: Using name field would be nice. wdyt?
```
steps:
  name: Validation step
  uses: ...
```
Same for other changes :)

sullis: Fixed.

sullis:
WDYT? @jan25

jan25:
Thanks @sullis. LGTM. 

Rest of yaml can follow similar form too. Probably can go in separate PR.

PR #769, opened 2020-07-21
fix(build): renable and reverse logic to prevent the accidental disabling of integration tests

fieldju:
This PR: https://github.com/spinnaker/kayenta/pull/708/files disabled integration tests and they have not been running since April, when I thought they had been.

This PR reverses the logic to make them opt-out rather than opt-in so they do not get accidentally disabled.

### DIFF
--- a/kayenta-integration-tests/kayenta-integration-tests.gradle
+++ b/kayenta-integration-tests/kayenta-integration-tests.gradle
@@ -16,8 +16,8 @@ test.testLogging {
 
 gradle.taskGraph.whenReady {
     tasks.test.enabled = (
-      properties.getOrDefault('ENABLE_INTEGRATION_TESTS', 'false') == 'true'
-        || System.getProperty('ENABLE_INTEGRATION_TESTS', 'false') == 'true'
-        || System.getenv().getOrDefault('ENABLE_INTEGRATION_TESTS', 'false') == 'true'
+      properties.getOrDefault('DISABLE_INTEGRATION_TESTS', 'false') != 'true'
+        || System.getProperty('DISABLE_INTEGRATION_TESTS', 'false') != 'true'
+        || System.getenv().getOrDefault('DISABLE_INTEGRATION_TESTS', 'false') != 'true'
     )
 }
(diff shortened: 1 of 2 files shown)


fieldju:
FYI, I expect this first build to fail, we run these again internally and noticed that they are broken between 1.20 and 1.21.

fieldju:
image

☝️ that's new between from one of these PRs

https://github.com/spinnaker/kayenta/pulls?q=is%3Apr+is%3Aclosed+label%3Atarget-release%2F1.21

Not sure if bug in Kayenta or maybe something wrong with test containers and the integration tests.

@Aloren if you could look, that would be great. I will attach a debugger and dig into the logs more tomorrow.

fieldju:
https://github.com/spinnaker/kork/pull/592
☝️ This pr introduced a micrometer registry bean which overrides the one created by the integration tests.

This PR overwrites that bean so the old logic works.

Aloren:
Hm.. This is interesting, but I am not sure that this is a proper fix for that or am I missing something?

fieldju:
> Hm.. This is interesting, but I am not sure that this is a proper fix for that or am I missing something?

I fixed it by adding that bean in the test config, just need to get this merged.

kayenta-integration-tests/kayenta-integration-tests.gradle
```diff
@@ -16,8 +16,8 @@ test.testLogging {
 
 gradle.taskGraph.whenReady {
     tasks.test.enabled = (
-      properties.getOrDefault('ENABLE_INTEGRATION_TESTS', 'false') == 'true'
```
kevinawoo, line 19: Ah I see. [github action build.yml:32](https://github.com/spinnaker/kayenta/pull/708/files#diff-898e737ee4e56ba042bda31764bed49eL32) and [github action pr.yml:27](https://github.com/spinnaker/kayenta/pull/708/files#diff-d9db351496552ae42395e3057e47624eL27) was removed, and with the old logic, it would have never ran.

Aloren:
> Fixing the fact that graphite is not using place holders when no data is returned is NOT fixed by this PR.
That is its own issue and PR and is orthogonal to the integration tests being disabled and not generating mock data properly.

@fieldju Thanks, that was exactly that I didn't understand :D

PR #776, opened 2020-07-30
Fix Canary Indefinitely Retrying

dwest-netflix:
This addresses a bug we discovered when running a config with no metrics. Orca would retry indefinitely waiting for non existent stages to complete.
Added a test on one of the controllers. Adding the check to the executionMapper seemed to be the safest spot since there seems to be an abundant number of ways to trigger a canary. 
Happy to adjust things. Just let me know.

### DIFF
--- a/kayenta-web/src/test/java/com/netflix/kayenta/controllers/BaseControllerTest.java
+++ b/kayenta-web/src/test/java/com/netflix/kayenta/controllers/BaseControllerTest.java
@@ -4,6 +4,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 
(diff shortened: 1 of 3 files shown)


dwest-netflix:
I seem to have broken the build. Looking into it.

kayenta-web/src/test/java/com/netflix/kayenta/controllers/BaseControllerTest.java
```diff
@@ -74,21 +83,41 @@ StorageServiceRepository storageServiceRepository(List<StorageService> storageSe
     }
 
     @Bean
-    StorageService storageService() {
-      StorageService mock = mock(StorageService.class);
```
dwest-netflix, line 78: Because of how this was wired up it was being shared across all tests. Swapped it to be MockBean which resets after every test

PR #777, opened 2020-08-01
chore(dockerfile): change deprecated maintainer instructor to label

KeisukeYamashita:
## What

As title. The `MAINTAINER` label is deprecated and now we should use `LABEL` instead.

## Ref

https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN tar -xf /tmp/workdir/kayenta-web/build/distributions/kayenta.tar -C /opt
 #
 FROM alpine:3.11
 
-MAINTAINER delivery-engineering@netflix.com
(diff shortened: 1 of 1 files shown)


.idea/gradle.xml
```diff
@@ -1,12 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
fieldju, line 1: Why did you delete this file, that doesn't seem like it should be part of this PR.

I do believe that @plumpy added all the IDE files explicitly.

KeisukeYamashita: @fieldju Sorry, I've deleted this by mistake. Recovered it 🙇

PR #784, opened 2020-08-18
chore(upgradeconfigbin): Upgrades configbin v1 to configbin v2.

dwest-netflix:
This change upgrades configbin to be compatible with configbinv2. 
It's unclear to me why this exists in the open source side at all (since configbin is not open sourced). 
But since it does, it seems right to upgrade it.


kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/service/ConfigBinRemoteService.java
```diff
@@ -21,22 +21,26 @@
...
 public interface ConfigBinRemoteService {
-  @GET("/{ownerApp}/{configType}")
-  String list(@Path("ownerApp") String ownerApp, @Path("configType") String configType);
 
-  @DELETE("/{ownerApp}/{configType}/{configName}")
+  @GET("/browse/{ownerApp}/{configType}?pageSize=1000&showDeleted=false")
+  String list(
+      @Path("ownerApp") String ownerApp,
+      @Path("configType") String configType,
+      @Query("nextPageId") String nextPageId);
```
dwest-netflix, line 29: API requires paging now

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -285,27 +285,49 @@ public void deleteObject(String accountName, ObjectType objectType, String objec
...
-      try {
-        List<String> ids =
-            kayentaObjectMapper.readValue(jsonBody, new TypeReference<List<String>>() {});
-
-        if (ids.size() > 0) {
-          return ids.stream().map(i -> metadataFor(credentials, i)).collect(Collectors.toList());
+      List<String> ids = new ArrayList<>();
+      boolean hasNext = true;
+      String pageId = null;
+      while (hasNext) {
```
dwest-netflix, line 291: Logic is largely the same here. There is now a response object that wraps the list of ids and the response can be paged.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/service/ConfigBinRemoteService.java
haleyw: Does this require a coordinated change from the UI then?

haleyw: Nevermind -- I see this is used by us internally rather than this being what Kayenta exposes to Spinnaker.

dwest-netflix: No change on the UI side. At start Kayenta will query for all configurations stored in ConfigBin. In this case we will pull the first 1000. We typically have 400 active configurations. But, in the event that we have over 1000, the ConfigBinStorageService will pull the next 1000 and combine results until no more pages are returned.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -285,27 +285,49 @@ public void deleteObject(String accountName, ObjectType objectType, String objec
...
+              response.nameVersions.stream().map(nv -> nv.configName).collect(Collectors.toList()));
+        } catch (IOException e) {
+          log.error("List failed on path {}: {}", ownerApp, e);
+          return Collections.emptyList();
         }
-      } catch (IOException e) {
-        log.error("List failed on path {}: {}", ownerApp, e);
+      }
+
+      if (ids.size() > 0) {
```
haleyw, line 313: NIT - I don't think you actually need this check...
If you stream an empty list of IDs, then you would just end up collecting an empty list. Doesn't hurt anything, but may simplify the branching.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -285,27 +285,49 @@ public void deleteObject(String accountName, ObjectType objectType, String objec
...
+                          MAX_RETRIES,
+                          RETRY_BACKOFF));
+          ConfigBinListResponse response =
+              kayentaObjectMapper.readValue(jsonBody, ConfigBinListResponse.class);
+          hasNext = response.hasNext;
+          pageId = response.nextPageId;
+          ids.addAll(
+              response.nameVersions.stream().map(nv -> nv.configName).collect(Collectors.toList()));
+        } catch (IOException e) {
+          log.error("List failed on path {}: {}", ownerApp, e);
```
haleyw, line 308: NIT - Unrelated to your change...I don't think will this actually log the full stack trace. I think this is going to call the toString() on the exception which will give the message rather than printing the stack trace. Stacktrace may be more useful for debugging.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -285,27 +285,49 @@ public void deleteObject(String accountName, ObjectType objectType, String objec
...
+      boolean hasNext = true;
+      String pageId = null;
+      while (hasNext) {
+        try {
+          final String currentPage = pageId;
+          String jsonBody =
+              AuthenticatedRequest.allowAnonymous(
+                  () ->
+                      retry.retry(
+                          () -> remoteService.list(ownerApp, configType, currentPage),
```
haleyw, line 298: Just checking it's ok for the currentPage to be null which it will be the first time through?

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
dwest-netflix: Nice catch. This is a hold over from the old code. I sometimes get blinders on when I'm doing something a bit more surgical.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
dwest-netflix: I'll fix this up

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
dwest-netflix: Yup in fact it's a necessity. Retrofit, if a Query param is null, will not add it to the request. Otherwise configbin would freak out about it being blank

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
dwest-netflix: So, surprisingly, it turns out that this line works. It ignores the final {} and outsputs something like ERROR: List failed on path kayenta: {}
(full exception stacktrace here) so I removed the extra {}

PR #796, opened 2020-09-07
fix(standalone-canary-analysis): set parent pipeline execution id into standalone canary analysis response

Aloren:
`CanaryAnalysisExecutionStatusResponse`'s `parentPipelineExecutionId` is always null.

Also simplified setting of other parameters to `CanaryAnalysisExecutionStatusResponse` using `map` method:
```
    public <U> Optional<U> map(Function<? super T, ? extends U> mapper) {
        Objects.requireNonNull(mapper);
        if (!isPresent()) {
            return empty();
        } else {
            return Optional.ofNullable(mapper.apply(value));
        }
    }
```
Since it already wraps nullable object into `Optional.ofNullable` there is no need to wrap it manually.

### DIFF
--- a/kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/service/CanaryAnalysisService.java
+++ b/kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/service/CanaryAnalysisService.java
@@ -160,8 +160,6 @@ public class CanaryAnalysisService {
    */
   protected CanaryAnalysisExecutionStatusResponse fromExecution(PipelineExecution pipeline) {
 
-    boolean isComplete = pipeline.getStatus().isComplete();
-    ExecutionStatus pipelineStatus = pipeline.getStatus();
     CanaryAnalysisExecutionStatusResponse.CanaryAnalysisExecutionStatusResponseBuilder
         responseBuilder =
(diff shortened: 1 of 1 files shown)


Aloren:
@fieldju could you please review? Thanks!

kayenta-standalone-canary-analysis/src/main/java/com/netflix/kayenta/standalonecanaryanalysis/service/CanaryAnalysisService.java
```diff
@@ -179,48 +177,44 @@ protected CanaryAnalysisExecutionStatusResponse fromExecution(PipelineExecution
...
-                .complete(isComplete)
-                .executionStatus(pipelineStatus);
+                .complete(pipeline.getStatus().isComplete())
+                .executionStatus(pipeline.getStatus());
 
     // Add the request and config info if possible
     pipeline.getStages().stream()
         .filter(stage -> stage.getType().equals(SetupAndExecuteCanariesStage.STAGE_TYPE))
         .findFirst()
+        .map(stage -> stage.getContext().get(CANARY_ANALYSIS_CONFIG_CONTEXT_KEY))
```
Aloren, line 187: If this is null, it will be wrapped into `Optional.ofNullable`

fieldju (approved):
LGTM,

Sorry about the long delay in the review.
Need to adjust my Gmail filters so that I get notifications again.

PR #797, opened 2020-09-09
feat(prometheus): add Prometheus health indicator (disabled by default)

Aloren:


Aloren:
Rebased onto master.
Added tests for `PrometheusResponseConverter` and `PrometheusRemoteService`.

PR #805, opened 2020-10-06
chore(java11): Target Java 11

cfieber:


.idea/compiler.xml
```diff
@@ -3,6 +3,70 @@
   <component name="CompilerConfiguration">
```
cfieber, line 3: oops will undo this ...

.idea/gradle.xml
```diff
@@ -1,11 +1,45 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
cfieber, line 1: and this..

kayenta-core/src/main/java/com/netflix/kayenta/service/MetricSetPairListService.java
```diff
@@ -35,7 +35,7 @@
   private final StorageServiceRepository storageServiceRepository;
 
   public List<MetricSetPair> loadMetricSetPairList(String accountName, String metricSetPairListId) {
-    String resolvedAccountName = getAccount(accountName);
+    var resolvedAccountName = getAccount(accountName);
```
cfieber, line 38: a `var` to ensure we JDK11

Dockerfile.compile
```diff
@@ -1,11 +1,9 @@
 FROM alpine:3.11
 RUN apk add --update \
     openjdk11 \
-    openjdk8 \
     && rm -rf /var/cache/apk
 LABEL maintainer="sig-platform@spinnaker.io"
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk
```
plumpy, line 6: I've been removing this `JAVA_HOME` too... there's only one JDK installed so we don't need it and it feels like one less thing to fix if we change distros or something like that...

plumpy (approved):
Awesome, thanks!

PR #811, opened 2020-10-09
chore(dependencies): update scala version to java 11 recommended

cfieber:


cfieber:
from https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html 2.12.12 is the recommended version with Java 11

csanden (approved):
LGTM

PR #819, opened 2020-11-24
feat(judge): add ability to mute metric result from total score computation

Aloren:
We are using canary analysis for lots of services internally in Playtika. Sometimes we need to add new metrics for the analysis and as it turns out -- we don't always start with ideal configuration and as a result it may lead to failed canary analysis. We want users to treat canary analysis as a real indicator of issues with deployment, so ideally we want to enable only the metrics that guarantee correct alarming 100% of the time. The `muted` property gives an ability to analyze metric, but skip it's result in a total score computation, so in case it is not yet perfect we won't affect deployments.

Implementation is backward compatible.

### DIFF
--- a/docs/canary-config.md
+++ b/docs/canary-config.md
@@ -77,8 +77,9 @@ Describes how to judge a metric, see the [Netflix Automated Canary Analysis Judg
 - `nanStrategy` (enum[string], optional) - How to handle NaN values which can occur if the metric does not return data for a particular time interval.
   - `remove` (default) - Use when you expect a metric to always have data and you want the NaNs removed from your data set (usage metrics).
   - `replace` - Use when you expect a metric to return no data in certain use cases and you want the NaNs replaced with zeros (for example: count metrics, if no errors happened, then metric will return no data for that time interval).
-- `critical` **true** (boolean, optional) - Use to fail the entire canary if this metric fails (recommended for important metrics that signal service outages or severe problems).
-- `mustHaveData` **true** (boolean, optional) - Use to fail a metric if data is missing.
+- `critical` **false** (boolean, optional) - Use to fail the entire canary if this metric fails (recommended for important metrics that signal service outages or severe problems).
+- `muted` **false** (boolean, optional) - Use to mute metric result in a total score computation.
+- `mustHaveData` **false** (boolean, optional) - Use to fail a metric if data is missing.
 - `effectSize` ([EffectSize](#effectsize), optional) - Controls how much different the metric needs to be to fail or fail critically.
 - `outliers` ([Outliers](#outliers), optional) - Controls how to classify and handle outliers.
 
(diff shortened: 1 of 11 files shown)


Aloren:
@ivanphdz @fieldju do I need to make some changes in the `provider-integration-tests` step to run my newly added test?

ivanphdz:
> @ivanphdz @fieldju do I need to make some changes in the `provider-integration-tests` step to run my newly added test?

Hi, do you want to include it on the github checks? you can use the include/exclude github feature to add a task for a specific provider. 

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations

```yaml
     strategy:
      matrix:
        provider: ['Prometheus', 'Graphite']
        test: ['canaryAnalysisIsSuccessful', 'canaryAnalysisIsFailed']
        include:
        - provider: 'Prometheus'
           test: 'mutedMetricsAreNotTakenIntoFinalScore'
```

Aloren:
Thanks @ivanphdz for your reply. I've just looked into the logs and I'm not sure what is the reason for parallelising the tests this way. Now it looks like adding more integration tests will require adding each new test to the gradle configuration. Am I right regarding this point or missing something? Also each test now setups the whole context (redis, minio, prometheus, graphite, kayenta) just to run one single test. Is that expected? For decreasing time of running tests I would suggest using retrospective analysis instead -- before running canary analysis tests wait 1 min until all metrics get published into the time-series db and after that run all tests in parallel. This way we can save both the resources and time. What do you think?

ivanphdz:
Thanks @Aloren , yeah, we need to add the new test case to the github config if the test case has a different name from the standard ones (IsSuccessful, IsFailed) or the new provider depending on the case, I think we can run tests by provider on parallel instead of every test case, so we would need to change the script only when a new provider is added.

Sure I like your idea we can give it a try and work together on this. let me know if you need help.

fieldju:
> I would suggest using retrospective analysis instead
That would only test that retrospective analysis is working.

I'd like to keep the tests we have that test end to end the real use case of running a real canary analysis.
aka that the wait stages all work together.

docs/canary-config.md
```diff
@@ -77,8 +77,9 @@ Describes how to judge a metric, see the [Netflix Automated Canary Analysis Judg
 - `nanStrategy` (enum[string], optional) - How to handle NaN values which can occur if the metric does not return data for a particular time interval.
   - `remove` (default) - Use when you expect a metric to always have data and you want the NaNs removed from your data set (usage metrics).
   - `replace` - Use when you expect a metric to return no data in certain use cases and you want the NaNs replaced with zeros (for example: count metrics, if no errors happened, then metric will return no data for that time interval).
-- `critical` **true** (boolean, optional) - Use to fail the entire canary if this metric fails (recommended for important metrics that signal service outages or severe problems).
-- `mustHaveData` **true** (boolean, optional) - Use to fail a metric if data is missing.
+- `critical` **false** (boolean, optional) - Use to fail the entire canary if this metric fails (recommended for important metrics that signal service outages or severe problems).
+- `muted` **false** (boolean, optional) - Use to mute metric result in a total score computation.
```
fieldju, line 81: What do you think about `precludeFromScore` vs `muted`?

docs/canary-config.md
```diff
@@ -90,11 +91,28 @@ See the [Netflix Automated Canary Analysis Judge] and [Mann Whitney Classifier]
 
 ### Properties
 
+- `measure` **meanRatio** (enum[string], optional) - Specifies how effect size is measured.
```
fieldju, line 94:
```suggestion
- `measure` **meanRatio** (enum[string], optional) - Defaults to meanRatio. Specifies how effect size is measured.
```

fieldju:
@Aloren I think this looks good, I had a minor suggestion on `measure` and a question on naming that I want your opinion on.

docs/canary-config.md
fieldju: `muted` isn't necessarily self-describing, I wonder if a name like `precludeFromScore` would lead to an easier to understand UX for peeps that not like digging through code or docs.

fieldju:
I've worked around this by adding a group that is weighted 0, but I like this more :)

fieldju:
Its orthogonal to this PR, but I like paralyzing just the classes, as it does read better.

image
image

@ivanphdz also we should be uploading the tests reports like we do with Rosco

fieldju (approved):

Aloren:
@fieldju Sorry, I was in vacation, so had not a chance to look into the comments. :) Do you still want me to make requested changes?

PR #824, opened 2020-12-17
feature(Kayenta Standalone): allow specifying a control offset in standalone kayenta

AbdulRahmanAlHamali:
Fixes [6239](https://github.com/spinnaker/spinnaker/issues/6239)


imosquera:
@AbdulRahmanAlHamali - Thanks for the PR! Any chance you can add a test for the changes you submitted?

fieldju:
@AbdulRahmanAlHamali 

I'd rather have this work like the `/canary` endpoint where a caller can specify the start time for the control and experiment directly rather than an offset for just the control.

Also its not clear what the offset even is, given the docs in that annotation, I can only see that its a subtractive offset by looking at the code.

I think I'd rather see

`controlStartTimeIso` and `experimentStartTimeIso` fields added to the scope object and have conditional logic where those fields are used / override the `startTimeIso` field that sets both.

PR #827, opened 2021-01-15
fix(azure): HTTPS support for Azure Storage (spinnaker#6205)

man0s:
**Why:**
- Kayenta has support for storing objects and metrics in Azure Storage but it's hardcoded to HTTP.
- HTTP is insecure.
- Azure Storage by default supports HTTPS.

**What:**
- Changed `DefaultEndpointsProtocol` to HTTPS

**Issue:**
- https://github.com/spinnaker/spinnaker/issues/6205


duftler (commented):
LGTM.

PR #837, opened 2021-03-18
feat: added hooks before archiving executions so that spring security…

fieldju:
… authentication details can be read from the pipeline context and injected into the security context in extension projects

See :point_down:  for example usage
image
image


link108 (approved):
LGTM 🎉  thank you 🙏

PR #844, opened 2021-05-12
fix(datadog): fix null handling in DatadogTimeSeries

fyi-coursera:
There was an attempted fix for this earlier (https://github.com/spinnaker/kayenta/pull/832) but it's incorrect. That fix caused every adjusted point to become `NaN` after the first `null` value. As an example, the `handlesNullInPointlist` test case would fail with `java.lang.AssertionError: expected:<[1.0, NaN, 1.0, 1.0, 1.0, NaN, NaN, NaN, NaN, NaN, 1.0]> but was:<[1.0, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN, NaN]>`


dreynaud (approved):

dreynaud:
Thanks for the fix @fyi-coursera ! 👍

PR #846, opened 2021-05-20
feat(config): Add simple server-side validation of metric names.

duftler:


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/6432.

kayenta-web/src/test/java/com/netflix/kayenta/controllers/CanaryConfigControllerTest.java
```diff
@@ -68,4 +72,20 @@ public void getCanaryConfig_returnsBadRequestResponseForNotResolvedAccount() thr
...
+  public void postCanaryConfig_returnsBadRequestResponseForDuplicateMetricName() throws Exception {
+    InputStream testConfig =
+        BaseControllerTest.class
+            .getClassLoader()
+            .getResourceAsStream("canary-configs/duplicated-metric-name.json");
+    String testConfigContent = IOUtils.toString(testConfig, "UTF-8");
+
+    this.mockMvc
+        .perform(post("/canaryConfig").contentType("application/json").content(testConfigContent))
+        .andDo(print())
```
anotherchrisberry, line 86: what does `print()` do?

anotherchrisberry:
One question but LGTM

anotherchrisberry (approved):

kayenta-web/src/test/java/com/netflix/kayenta/controllers/CanaryConfigControllerTest.java
duftler: It just dumps out details about the response. In response to your comment, I tried the same test without it (I had it there because I wanted to fit in with the neighbors). The details about the request/response are duplicated when the line is present, and appear just once when the line is removed.

Then I found this: https://github.com/spinnaker/kayenta/blob/master/kayenta-web/src/test/java/com/netflix/kayenta/controllers/BaseControllerTest.java#L71

Have updated this test and its siblings to remove the explicit `.andDo(print())` calls.

PR #853, opened 2021-11-09
fix(graphite): urlencode target query

perek:
We noticed on newer versions of graphite and certain queries that the target parameter must be properly urlencoded before being sent to the graphite service via GET. Previously we had set this to false as very old versions of graphite did not handle the url encoding properly, I tested this with our graphite and the change seems safe and fixes the bugs we saw in newer graphites.


mattgogerly:
Makes sense to me. https://github.com/prometheus-junkyard/promdash/issues/280#issuecomment-70497298

mattgogerly (approved):

PR #864, opened 2022-02-02
fix(timeouts): Enable timeout configs on retrofit

jasonmcintosh:
Adds read & connect timeouts to increase from default 10 seconds Ok http client uses.  Probably need to adjust:
https://github.com/spinnaker/kayenta/blob/master/kayenta-core/src/main/java/com/netflix/kayenta/retrofit/config/RetrofitClientFactory.java
to ALSO support these settings...


jasonmcintosh:
Turns out this uses Kork's config I just missed the line :). The authenticated retrofit client on the other side does NOT seem ot use it... separate PR for that will be incoming...

PR #875, opened 2022-03-05
fix(image): Add tzdata to set timezone on kayenta pod

abhinaybyrisetty:
Issue:
Inorder to set timezone on a container, setting it through "TZ" environment variable is one of the ways. However, as tzdata package is not installed on the pod, setting "TZ" environment variable doesn't make any difference.

Solution:
Having "tzdata" package installed on the pod gives the flexibility to the user to simply set "TZ" env var and configurable timezone of the pod for better troubleshooting of the logs.


link108:
closing, see https://github.com/spinnaker/gate/pull/1525 for more details

PR #888, opened 2022-04-27
fix(ci): fetch previous tag from git instead of API

kskewes-sf:
Previously the release info script would query the GitHub HTTP API and retrieve
the latest "Release" (git tag) published.
The latest "Release" could (likely) be for a different `release-*` branch or
even `master` rather than the previous tag on the same branch as the new
tag.
This resulted in "Release" changelogs containing commits that dont
exist in the tags history.

By instead querying git itself we can fetch the previous tag in history
and then generate an accurate changelog, eg: between `v1.7.4` and
`v1.7.3` rather than potentially `v1.7.4` and `v1.10.0`.

Run script through `shellcheck` and `shfmt`.

### DIFF
--- a/.github/workflows/release_info.sh
+++ b/.github/workflows/release_info.sh
@@ -1,20 +1,24 @@
 #!/bin/bash -x
 
-# Only look to the latest release to determine the previous tag -- this allows us to skip unsupported tag formats (like `version-1.0.0`)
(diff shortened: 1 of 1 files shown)


dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.27.x release-1.28.x

PR #926, opened 2022-12-19
chore(cleanup): Doing refactoring of Kayenta to cleanup codebase

jasonmcintosh:
There are several long term goals here, this is step "A"
* MAIN long term goal is to support property refreshes via RefreshScoped/Configproperties (aka kork-config) and dynamic accounts.  Doing cleanup first.
* Get ride of "METRICS_STORE", "CONFIG_STORE", "OBJECT_STORE" - support a SQL storage engine instead, and all other accounts are metric stores.
* AND Make the account repo eventually usable as a REAL repo (e.g. support CRUD operations & storage of accounts in a real repo)
Before doing the above, trying to refactor the codebase a bit to be a touch cleaner by:
* Merge the "subtypes" e.g. "s3" into "aws", "stackdriver" and "gcs" into "google", "blobs" into azure.  It's a bit TOO broken down (opinion here, not required, but when doing tests, made it a LOT less... cumbersome to identify resource dependencies)
* Remove duplication/handling of account information.  Start with the same pattern as prometheus.  Remove the duplicate code(where an account properties are loaded, copied to a nearly duplicate account class with "credentials" removed).  Instead, leave those as json ignored fields or transients to prevent serialization.  This DRASTICALLY simplifies both load and config and account management code.
* Eventually change a few more of the APIs that have some rather... bad handling of account repo stuff.

**Additional PRs with more refactoring will be coming.**  

Note some pieces that are POTENTIALLY a concern:
* We moved to "name" across ALL accounts as an Equals handler.  Currently there's nothing that prevents duplicate account names across resource types (you can have a "metrics-account" in GCP and Prometheus for example and it likely would work... until you make API calls).
* Changing the account load handling... if something really IS persisting these or serializing the configs someplace, that may cause some issues as we shrink things down to simpler code
* Hashcode on "name" ... that's for sure equals, but not sure hashcode for that makes sense, yet... not sure it doesn't.  Frankly I doubt the hashcode handling is used many places other than for some lookups.


kayenta-core/src/main/java/com/netflix/kayenta/security/AccountCredentials.java
```diff
@@ -17,39 +17,44 @@
...
    * used locations, but the full list is unknown by the metrics service.
    */
-  default List<String> getRecommendedLocations() {
-    return Collections.emptyList();
-  }
+  @Builder.Default private List<String> recommendedLocations = Collections.emptyList();
 
   @JsonIgnore
-  T getCredentials();
+  public abstract T getCredentials();
```
jasonmcintosh, line 55: Note, this "getCredentials" should go away once all refactoring is done, as it'll have no meaning.  The account itself is the "Credentials" to communicate with a remote source.

jvz (approved):
LGTM

kayenta-core/src/main/java/com/netflix/kayenta/security/AccountCredentials.java
jvz: I assume this was probably copied from Clouddriver originally where three levels of indirection are needed for reasons.

kayenta-core/src/main/java/com/netflix/kayenta/security/AccountCredentials.java
```diff
@@ -17,39 +17,44 @@
...
 
-public interface AccountCredentials<T> {
-  String getName();
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+@AllArgsConstructor
+@EqualsAndHashCode(of = "name")
+public abstract class AccountCredentials<T> {
```
jvz, line 32: Good thing about making this abstract is that it'll fit more naturally as an implementation of `Credentials` from Kork. Considering that I've been consolidating a lot of account management stuff into Kork lately, this will be good.

jasonmcintosh:
@Spinnakerbot add-label ready-to-merge

PR #935, opened 2023-02-28
chore(cleanup): refactoring down to single accounts across all services

jasonmcintosh:
Significant refactor of accounts to reduce duplication.  Prepares for SQL account repository handling.


kayenta-atlas/src/main/java/com/netflix/kayenta/atlas/config/AtlasManagedAccount.java
```diff
@@ -16,21 +16,35 @@
...
 
-  private List<String> recommendedLocations;
+  private List<String> recommendedLocations = new ArrayList<>();
+
+  @Override
+  public List<String> getLocations() {
+    return getBackendUpdater().getBackendDatabase().getLocations();
+  }
+
+  @JsonIgnore private transient BackendUpdater backendUpdater;
```
jvz, line 42: I know it might be a little early to ask, but if you intend on making these credentials storable, then it would also make sense for them to work similarly through the API. Simply supporting JPA isn't enough; we'd need a way to marshal and unmarshal these through the REST API so that they can be specified from somewhere other than a config file in the first place.

kayenta-influxdb/src/main/java/com/netflix/kayenta/influxdb/config/InfluxDbManagedAccount.java
```diff
@@ -16,19 +16,30 @@
...
+@Getter
+@Setter
+@SuperBuilder
+@NoArgsConstructor
+public class InfluxDbManagedAccount extends AccountCredentials<InfluxDbManagedAccount> {
   private String apiKey;
   private String applicationKey;
 
   @NotNull private RemoteService endpoint;
+  @JsonIgnore private transient InfluxDbRemoteService influxDbRemoteService;
```
jvz, line 39: Might make more sense to use `@JacksonInject` here instead?

jvz:
Looks interesting, but this diverges so much from anything remotely similar to the Kork credentials API that I'm still unsure how this will eventually integrate with secrets, storage, and the REST API. I see a note about some code being potentially useful as a `CredentialsParser`, but that is an API for turning a `CredentialsDefinition` into a `Credentials` object, and this PR seems to combine those two concepts into the one `AccountCredentials` class. Given the inability to update third party classes to implement `Credentials`, it seems as though there may be a gap here?

jvz:
I see. We've diverged quite a bit then at this point, so my previous RFC to add RBAC to Kayenta won't be possible until we've figured that out.

dbyron-sf:
Any chance this can wait til after 1.30?  🤞 that happens "real soon now" ™️

PR #938, opened 2023-03-17
feat: SPLAT-569: Add SQL data source for account credentials and account configurations persistence.

armory-abedonik:
**How to use?**

SQL is disabled by default, use these properties (in kayenta-local.yml) to enable MySQL or PostreSQL data source.

```
kayenta:
  sql:
    enabled: true

spring:
  datasource:
    driver-class-name: com.mysql.cj.jdbc.Driver or org.postgresql.Driver
    username: ...
    password: ...
    url: jdbc:mysql://...
```


kayenta-sql/src/main/java/com/netflix/kayenta/sql/storage/SqlStorageService.java
```diff
@@ -0,0 +1,229 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Singular;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Service;
+
+@Builder
+@Service
+@ConditionalOnProperty("kayenta.sql.enabled")
+public class SqlStorageService implements StorageService {
```
jasonmcintosh, line 48: I'd LIKE to get some simple tests on this please :)

jasonmcintosh (changes requested):

kayenta-sql/src/main/resources/db/changelog/202303091551-create-canary-archive-table.yml
```diff
@@ -0,0 +1,29 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: VARCHAR(255)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: content
+                  type: TEXT
```
jvz, line 17: Ideally, this column type should be `JSON` for MySQL-style databases and `JSONB` for PostgreSQL-style databases. This makes it much easier to handle data migrations as an admin over time, and on the JSONB side of things, it's far more space efficient. Same for the other tables with a generic column for holding JSON data.

kayenta-sql/src/main/resources/db/changelog/202303091551-create-canary-archive-table.yml
```diff
@@ -0,0 +1,29 @@
+databaseChangeLog:
+  - changeSet:
+      id: create-canary-archive-table
+      author: alexey.bedonik
+      changes:
+        - createTable:
+            tableName: canary_archive
```
jvz, line 7: For this table (and the rest), it may be helpful to set some MySQL settings since it defaults to bullshit. You can use something like this in your create table statement:

```
    - modifySql:
        dbms: mysql
        append:
          value: " engine innodb DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_unicode_ci"
```

kayenta-sql/src/main/resources/db/changelog/202303091551-create-canary-archive-table.yml
jvz: I have an example in https://github.com/spinnaker/kork/blob/ec5c486656cd2ba28c50cc1c11f6e83879b50a4f/kork-credentials-sql/src/main/resources/db/changelog/20230217-accounts-body-typefix.yml on how I was able to finagle Liquibase to support this. If you default to a `JSON` column type with a followup changelog entry to update PostgreSQL databases to use `JSONB`, that seems to work. Note that these columns work effectively like `TEXT` columns when doing normal queries, so there's nothing to change in your entities.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/storage/model/SqlMetricSetPairs.java
```diff
@@ -0,0 +1,34 @@
...
+import java.time.Instant;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import lombok.Data;
+
+@Data
+@Entity
+@Table(name = "metric_set_pairs")
+public class SqlMetricSetPairs {
```
jasonmcintosh, line 28: SO one thing:  IDEALLY these all should extend a parent class that defines the createdAt, updatedAt, content, etc. to reduce duplication.

jasonmcintosh: Arguably... these could all be the SAME object with a "type" reference in the database table to do uniqueness.

jasonmcintosh (approved):
I'm good with this at it is now can refactor/change behaviors later.  It's a start and behind feature flags.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/storage/model/SqlMetricSetPairs.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.sql.storage.model;
+
+import javax.persistence.Table;
+
+@Table(name = "metric_set_pairs")
```
jvz, line 21: Nit: table names should typically be singular as they are describing the contents of a row within the table. You're already following that naming convention in the rest of your tables/entities.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/storage/model/SqlMetricSets.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.kayenta.sql.storage.model;
+
+import javax.persistence.Table;
+
+@Table(name = "metric_sets")
```
jvz, line 21: Nit: this could also be singular.

kayenta-sql/src/main/resources/db/changelog/202303091555-adjust-tables-for-postres.yml
```diff
@@ -0,0 +1,25 @@
+databaseChangeLog:
+  - changeSet:
+    id: adjust-tables-for-postgres
```
jvz, line 3: Thanks! ❤️

jvz (approved):
Generally looks good. Only had a couple nit-picks that aren't blockers.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/storage/model/SqlMetricSetPairs.java
armory-abedonik: Yes, each row in the table will contain `metric_set_pairs`. Plural form in the given case is  `metric_set_pairs_items`

kayenta-sql/kayenta-sql.gradle
```diff
@@ -0,0 +1,7 @@
+dependencies {
+    implementation project(":kayenta-core")
+    implementation "org.springframework.boot:spring-boot-starter-data-jpa"
+    implementation "org.liquibase:liquibase-core:4.20.0"
+    implementation "mysql:mysql-connector-java:8.0.25"
+    implementation "org.postgresql:postgresql"
```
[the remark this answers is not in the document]
jasonmcintosh: Probably unless using testContainers to test mysql/postgresql integration?  Shouldn't impact things, but would help.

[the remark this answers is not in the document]
jvz: TestContainers downloads containers on demand. The runtime only dependency type should work with tests, too.

PR #940, opened 2023-03-24
feat: SPLAT-582: Add a storage data migrator.

armory-abedonik:
**How to use?**

Data migration and SQL is disabled by default, use these properties (in kayenta-local.yml) to enable the data migration and MySQL or PostgreSQL data source.

```
kayenta:
  sql:
    enabled: true
  data-migration:
    enabled: true
    sourceStorageServiceClassName: com.netflix.kayenta.s3.storage.S3StorageService
    targetStorageServiceClassName: com.netflix.kayenta.sql.storage.SqlStorageService
    sourceAccountName: monitoring
    targetAccountName: monitoring
  aws:
    enabled: true
    accounts:
      - name: monitoring
        bucket: ...
        region: us-west-2
        explicitCredentials:
          accessKey: ...
          secretKey: ...
          sessionToken: ...
        rootFolder: kayenta
        roleName: default
        supportedTypes:
          - OBJECT_STORE
          - CONFIGURATION_STORE
          - METRICS_STORE
  cloudwatch:
    enabled: true
  s3:
    enabled: true

spring:
  datasource:
    driver-class-name: com.mysql.cj.jdbc.Driver or org.postgresql.Driver
    username: ...
    password: ...
    url: jdbc:mysql://...
```

### DIFF
--- a/kayenta-sql/src/main/java/com/netflix/kayenta/sql/config/SqlConfiguration.java
+++ b/kayenta-sql/src/main/java/com/netflix/kayenta/sql/config/SqlConfiguration.java
@@ -16,17 +16,9 @@
 
 package com.netflix.kayenta.sql.config;
 
-import com.netflix.kayenta.sql.migration.StorageDataMigrator;
-import com.netflix.kayenta.storage.StorageService;
-import java.io.IOException;
-import java.util.List;
-import java.util.concurrent.Executors;
-import lombok.RequiredArgsConstructor;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
-import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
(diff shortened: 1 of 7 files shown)


jasonmcintosh:
Will look more here later... my INITIAL thoughts are... the migrator doesn't seem SQL specific, and... i'd love it if "accounts" as a concept for STORAGE was removed.  that the metric account was a field on any persisted set of metric data, and permissions are set on 'metric accounts' kinda things.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/config/SqlConfiguration.java
```diff
@@ -28,4 +36,35 @@
...
+
+  private final SqlProperties sqlProperties;
+  private final List<StorageService> storageServices;
+
+  @Bean
+  @ConditionalOnProperty("kayenta.sql.migration.enabled")
+  public StorageDataMigrator storageDataMigrator() throws IOException {
+    var sourceStorageServiceClassName =
+        sqlProperties.getMigration().getSourceStorageServiceClassName();
+    var targetStorageServiceClassName =
```
jasonmcintosh, line 51: Given SQL storage doesn't have a concept of "account".  I'm inclined to remove the idea of a target account.    OR make this migrator more generic and independent of SQL config and let it copy data from one storage account to another of any kind.  

SQL - I'd like to remove "accounts" entirely.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/migration/StorageDataMigrator.java
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+
+    for (var objectKey : objectKeysToMigrate) {
+      executorService.submit(
+          () -> {
+            try {
+              var object =
+                  sourceStorageService.loadObject(sourceAccountName, objectType, objectKey);
+              targetStorageService.storeObject(targetAccountName, objectType, objectKey, object);
```
jasonmcintosh, line 87: Given this re-runs regularly, I'd expect this to output a LOT of data after a few cycles.  I'm wondering if there's a way to compare time stamps or something else so we don't spam logs once this is enabled.  There could be a lot of data here.

kayenta-sql/src/main/java/com/netflix/kayenta/sql/migration/StorageDataMigrator.java
```diff
@@ -0,0 +1,110 @@
...
+                  "Unable to migrate objectType: {}, objectKey: {}", objectType, objectKey, e);
+            }
+          });
+    }
+  }
+
+  @Scheduled(fixedDelay = 60000)
+  public void migrate() {
+    if (!sqlProperties.getMigration().isEnabled()) {
+      log.info("Migration is disabled");
```
jasonmcintosh, line 99: Given this class should ONLY load if migration is enabled... isn't this somewhat pointless?

jasonmcintosh (approved):

PR #957, opened 2023-04-13
fix(signalfx): Fixed metric type missing due to duplicated field from parent class

jasonmcintosh:
When we refactored the code to extend a base class... we missed a spot that had overloaded the type.  As a result of this no types were returned.  Drop downs for account lists currently don't work unless a type is shown for the given account.

### DIFF
--- a/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/security/SignalFxNamedAccountCredentials.java
+++ b/kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/security/SignalFxNamedAccountCredentials.java
@@ -21,9 +21,10 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.netflix.kayenta.retrofit.config.RemoteService;
 import com.netflix.kayenta.security.AccountCredentials;
 import com.netflix.kayenta.signalfx.service.SignalFxSignalFlowRemoteService;
-import java.util.List;
 import javax.validation.constraints.NotNull;
-import lombok.*;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 import lombok.experimental.SuperBuilder;
 
 @Getter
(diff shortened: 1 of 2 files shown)


dbyron-sf:
@jasonmcintosh can you link the PR that introduced this bug please?

jasonmcintosh:
https://github.com/spinnaker/kayenta/commit/48348530b68930e7aea6f639e38e3d017fedd9de#diff-9641f219299469e2a68cad79922e76a58e6ad5ac7e55626106ac7c25d85703b3

Specifically when I refactored to a base account credentials:
https://github.com/spinnaker/kayenta/pull/926

dbyron-sf:
Looks like it was https://github.com/spinnaker/kayenta/pull/926, first introduced in 1.30.0.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/security/SignalFxNamedAccountCredentials.java
```diff
@@ -34,8 +35,6 @@ public class SignalFxNamedAccountCredentials extends AccountCredentials<SignalFx
 
   @NotNull private String name;
 
-  @NotNull @Singular private List<Type> supportedTypes;
```
dbyron-sf, line 37: This is the fix because https://github.com/spinnaker/kayenta/pull/926/files#diff-4b56b9f07abd19a8b5db2b64409e10bad9f639f0a1764e53934f97c03eb83caeR34 put `supportedTypes` in the base AccountCredentials class.

jasonmcintosh: Yep! I removed it from all the other accounts, apparently somehow missed signalfx.  I added a test, checked elsewhere couldn't find it elsewhere.  Got a test for it here, but long term if I can nuke this whole "copy the one into the other" code a lot of this'll go away.

dbyron-sf (approved):

jasonmcintosh:
https://github.com/Mergifyio backport 1.30.x

PR #959, opened 2023-04-24
fix: Errors handling and objects fetching from s3.

armory-abedonik:


jasonmcintosh (changes requested):
Need some tests please ;)

PR #972, opened 2023-06-23
feat(exceptions): Add SpinnakerRetrofitErrorHandler and replace RetrofitError catch blocks

abe21412:
This pull request registers SpinnakerRetrofitErrorHandler with each Retrofit.RestAdapter and replaces each RetrofitError catch block with a catch-block using SpinnakerServerException or the appropriate subclass. 

This change does not alter any of this service's behavior, it merely allows error messages to be surfaced even when the error was thrown in a microservice more than one network call from the service in which the request originated.

This is part of an effort to consume SpinnakerRetrofitErrorHandler in each Spinnaker microservice, as detailed in https://github.com/spinnaker/spinnaker/issues/5473


kayenta-atlas/kayenta-atlas.gradle
```diff
@@ -8,4 +8,6 @@ dependencies {
   api "com.squareup.okhttp:okhttp-apache"
 
   api "org.apache.commons:commons-io:1.3.2"
+
+  api "io.spinnaker.kork:kork-retrofit"
```
dbyron-sf, line 12: implementation?

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
```diff
@@ -88,10 +88,8 @@ public String executeQuery(
             metricsService.queryMetrics(
                 metricsAccountName, canaryConfig, canaryMetricConfig, canaryScope);
         success = true;
-      } catch (RetrofitError e) {
-
-        boolean retryable = isRetryable(e);
-        if (retryable) {
+      } catch (SpinnakerHttpException e) {
+        if (e.getRetryable()) {
```
dbyron-sf, line 92: Pretty tempting to use [RetrySupport](https://github.com/spinnaker/kork/blob/44281f9c707997b9942185f625637f1efb9bfcc6/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/core/RetrySupport.java) here.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -87,7 +89,7 @@ public <T> T loadObject(String accountName, ObjectType objectType, String object
                       () -> remoteService.get(ownerApp, configType, objectKey),
                       MAX_RETRIES,
                       RETRY_BACKOFF));
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       throw new NotFoundException("No such object named " + objectKey);
```
dbyron-sf, line 93: Proabably a followup PR, but is it really appropriate to throw a NotFoundException in all cases here?  I would have though it only applies for a SpinnakerHttpException with a 404 response code...but in that case the only reason to catch and re-throw could be to add a more descriptive error message
```
throw e.newInstance("No such object named " + objectKey);
```
might cover it.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
```diff
@@ -337,7 +339,7 @@ private Map<String, Object> metadataFor(ConfigBinNamedAccountCredentials credent
                       () -> remoteService.get(ownerApp, configType, id),
                       MAX_RETRIES,
                       RETRY_BACKOFF));
-    } catch (RetrofitError e) {
+    } catch (SpinnakerServerException e) {
       throw new IllegalArgumentException("No such object named " + id);
```
dbyron-sf, line 343: Similar question here as https://github.com/spinnaker/kayenta/pull/972/commits/c75541ca7a923355aa651275b6c8b9229b27f1e5#r1240412522, though IllegalArgumentException results in a 400 response, so `e.newInstance` may not work without other changes.

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
```diff
@@ -144,8 +144,9 @@ public List<MetricSet> queryMetrics(
       signalFlowExecutionResult =
           signalFlowService.executeSignalFlowProgram(
               accessToken, startEpochMilli, endEpochMilli, stepMilli, maxDelay, immediate, program);
-    } catch (RetrofitError e) {
-      ErrorResponse errorResponse = (ErrorResponse) e.getBodyAs(ErrorResponse.class);
+    } catch (SpinnakerServerException e) {
```
dbyron-sf, line 147: This is actually a change/reduction in functionality (olnly the message property instead of all ErrorResponse fields).  There's a plan for SpinnakerServerException (or perhaps it'll be only SpinnakerHttpException) to have an accessor for json response bodies as a Map, so perhaps it's better to wait for that?

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
abe21412: I agree but I think it makes more sense to save that for a follow-up PR since it's separate from replacing instances of `RetrofitError`

kayenta-signalfx/src/main/java/com/netflix/kayenta/signalfx/metrics/SignalFxMetricsService.java
abe21412: I'll revert this change and add a TODO with the note to replace RetrofitError once the Spinnaker exceptions have a response body accessor

abe21412: thanks, I'll check it out and make the necessary changes

abe21412: it looks like kayenta pulls in kork through its orca dependency unless I'm mistaken. I see that orca's main branch has the latest kork version with the accessor present but there hasn't been a release of orca with that change yet.

dbyron-sf: Aaah, right you are.  Lemme tag orca so kayenta picks up the new stuff.

dbyron-sf: And finally with https://github.com/spinnaker/kayenta/pull/974, kayenta gets the new stuff.

kayenta-objectstore-configbin/src/main/java/com/netflix/kayenta/configbin/storage/ConfigBinStorageService.java
dbyron-sf: That link doesn't take me to a question anymore.  This change seems OK to me as I'm looking at it now.  I do wonder if it would be an improvement to add e as the cause, like
```suggestion
      throw new IllegalArgumentException("No such object named " + id, e);
```

kayenta-core/src/main/java/com/netflix/kayenta/metrics/SynchronousQueryProcessor.java
```diff
@@ -146,28 +164,15 @@ private long getBackoffPeriodMs(int retryAttemptNumber) {
...
-    if (isNetworkError(e)) {
-      // retry in case of network errors
-      return true;
-    }
-    if (e.getResponse() == null) {
-      // We don't have a network error, but the response is null. It's better to not retry these.
-      return false;
-    }
-    HttpStatus responseStatus = HttpStatus.resolve(e.getResponse().getStatus());
+  private boolean isRetryable(SpinnakerHttpException e) {
```
dbyron-sf, line 167: Wouldn't it be simpler to do
```suggestion
  private boolean isRetryable(SpinnakerServerException e) {
```
and then keep an isNetworkError that also took a SpinnakerServerException argument?

abe21412: yeah, I think I see a way to do that. I'll update my branch

dbyron-sf (approved):

PR #1034, opened 2024-05-15
refactor(retrofit): Replace the RetrofitExceptionHandler with SpinnakerServerExceptionHandler

j-sandy:
As RetrofitExceptionHandler have been dismantled from [orca](https://github.com/spinnaker/orca/pull/4716), now replacing it with SpinnakerServerExceptionHandler.


dbyron-sf (changes requested):
Let's add SpinnakerServerExceptionHandler in one step, then tag/bump orca, then remove RetrofitExceptionHandler in a subsequent step.

dbyron-sf:
Included in https://github.com/spinnaker/kayenta/pull/1036

PR #1047, opened 2024-08-06
fix(stackdriver): handle null timeSeries and empty points

shlomodaari:
Adding an if statement to check if timeSeries is null and if the points are empty + tests


edgarulg (approved):

edgarulg:
@Mergifyio backport release-1.32.x release-1.33.x release-1.34.x release-1.28.x

PR #1059, opened 2024-09-05
Add api token capability to InfluxDB

belldina:
Added plumbing to read apiKey from Kayenta InfluxDB and wire it to a "Token <apiKey>" auth header for InfluxDB API requests.
Understood that it's not elegant to simply append more args to the createClient method, but that whole piece needs a rewrite to account for various auth methods, so I'm simply tacking on for now.


belldina:
is there anything else on my side that needs to be done?  I've squashed all commits and reformatted the commit title.

belldina:
updated the PR title to match the requests commit title format.

jasonmcintosh:
Hey sorry hadn't seen this :) HIGHLY recommend tagging people in slack - not all of us watch all the repos, though hopefully with monorepo that'll come soonish.  IF can update this, I can see about reviewing & merging!

PR #1065, opened 2024-11-15
chore(java): Full Java 17 support only

jasonmcintosh:


jasonmcintosh:
To note I DID NOT upgrade the base OS or the JDK to the JRE or anything on this PR.  I can do that if we wish...

PR #1086, opened 2025-04-21
Enable dynamic port allocation for Prometheus and management server during integration tests

rahul-chekuri:
- Renamed `EmbeddedPrometheusBootstrapConfiguration` as `TestPrometheusConfiguration` and now `TestPrometheusConfiguration` dynamically allocate and register Prometheus container ports without using hardcoded ports.

- Removed dependency on predefined `server.port` and `management.server.port` values for integration tests.

- Management server port is dynamically obtained during runtime and exposed for Prometheus to scrape metrics.

- Prometheus container’s mapped port is dynamically registered into the environment (`embedded.prometheus.port`) at runtime.

- Fixed `ManagementTest` to use the dynamically registered Prometheus port (`embedded.prometheus.port`).

- Added logic in `BaseIntegrationTest` to update `dynamicEndpoint` to `PrometheusRemoteService`


kayenta-integration-tests/src/test/resources/application-base.yml
```diff
@@ -28,8 +28,8 @@ kayenta:
   s3:
     enabled: true
 
-server:
-  port: 8080
+#server:
```
dbyron-sf, line 31: remove commented out code?

